PR #21, opened 2019-11-01
Use raw sql to change the backfill instance states.

mpawliszyn:
Otherwise we get optimistic lock exceptions due to the running instances
refreshing their leases.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/DbBackfillRun.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/DbBackfillRun.kt
@@ -133,12 +133,14 @@ class DbBackfillRun() : DbUnsharded<DbBackfillRun>, DbTimestampedEntity {
     // State can't be changed after being completed.
     checkState(this.state != BackfillState.COMPLETE)
     this.state = state
-    instances(session, queryFactory)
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/app/cash/backfila/service/DbBackfillRun.kt
```diff
@@ -133,12 +133,12 @@ class DbBackfillRun() : DbUnsharded<DbBackfillRun>, DbTimestampedEntity {
...
-        .forEach { instance ->
-          if (instance.run_state != BackfillState.COMPLETE) {
-            instance.run_state = state
-          }
-        }
+    // Set the state of all the instances that are not complete
+    val query = session.hibernateSession.createQuery("update DbRunInstance set run_state = :newState where backfill_run_id = :runId and run_state <> :completed")
+    query.setParameter("runId", id)
+    query.setParameter("newState", state)
+    query.setParameter("completed", BackfillState.COMPLETE)
```
shellderp: we should increment version!

mpawliszyn: Done

PR #31, opened 2019-11-06
Adds a config for the number of backfill runner threads per process.

mpawliszyn:


service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
```diff
@@ -53,7 +53,7 @@ class BackfilaServiceModule(
 
   @Provides @ForBackfilaScheduler @Singleton
   fun backfillRunnerExecutor(): ListeningExecutorService {
-    return MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(40,
+    return MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(config.backfill_runner_threads?: 40,
```
shellderp: you could add a default value to the config constructor instead

PR #35, opened 2019-11-18
Trying to fix proto problems.

mpawliszyn:
All the client protos should be in the client package as far as I can
tell.


client-misk/build.gradle
```diff
@@ -2,14 +2,6 @@ apply plugin: 'com.squareup.wire'
 apply plugin: 'kotlin'
```
adrw: For the `src/main/proto` path not existing compile error, you'll need to remove the wire plugin from `client-misk`

PR #40, opened 2019-12-13
Track Wire's new configuration names

swankjesse:
Make sure our MPP workaround works whatever Wire calls its stuff


swankjesse:
💙

PR #45, opened 2020-01-06
Backfila web UI: navbar color reflects environment

### LINKED ISSUE #16: Fix navbar not showing staging environment banner

zcross:
Fixes #16 

I noticed this using backfila. IMO, visual cues about the current environment can help reduce the risk of operational tasks, so this is pretty important.

That said, I kept the specific shade of red that we must've thought was nice when initially writing this.

Meta: The current README's instructions reference a nonexistent `BackfilaService.kt` _and_ there are some DI exceptions related to the web actions when trying to run `BackfilaDevelopmentService.kt`

### DIFF
--- a/service/web/tabs/app/src/utilities/theme.ts
+++ b/service/web/tabs/app/src/utilities/theme.ts
@@ -14,10 +14,7 @@ const environmentColorLookup: IEnvironmentToColorLookup = {
   PRODUCTION: color.red
 }
 
-export const backfilaRed = "#db2636"
-
 export const backfilaTheme: ITheme = {
   ...defaultTheme,
-  environmentToColor: environmentToColor(environmentColorLookup),
-  navbarBackground: backfilaRed
+  environmentToColor: environmentToColor(environmentColorLookup)
 }


shellderp:
does this still set the background to red/something else in staging? or is it foreground now?

meta: Yeah I have a fix locally for the dev stuff, will send a PR eventually

zcross:
No, this should only make production red. Staging _should_ be green!

service/web/tabs/app/src/utilities/theme.ts
```diff
@@ -11,13 +11,10 @@ const environmentColorLookup: IEnvironmentToColorLookup = {
...
-  PRODUCTION: color.red
+  PRODUCTION: "#db2636"
 }
 
-export const backfilaRed = "#db2636"
-
 export const backfilaTheme: ITheme = {
   ...defaultTheme,
-  environmentToColor: environmentToColor(environmentColorLookup),
-  navbarBackground: backfilaRed
```
adrw: Have you tested this locally? navbarBackground currently applies across all environments so this change would remove the red from the navbar even in production. Making navbar background environment specific is easily possible but would require associated Misk-Web changes.

zcross: Ok, I'm fine dropping the "backfilaRed" I think the stylistic quirkiness of that is less important than accurately reflecting environments as a visual cue

adrw: Instead of the suggested changes, can you remove the theme override in LayoutContainer and leave this file as is? https://github.com/cashapp/backfila/blob/229b39eddc60d6b0f4d735ffebb177f061bb21bf/service/web/tabs/app/src/containers/LayoutContainer.tsx#L19

zcross: Just changed this to use the default misk-web behavior w.r.t. navbarBackground.
My read of misk-web is that it will use `environmentToColor` as defined by the theme:

```
packages/@misk/core/src/features/Navbar/Banner.tsx
24:  background-color: ${theme.environmentToColor(environment)} !important;
```

zcross: Gonna decline this since there's a potentially better fix in misk-web that will obviate changes in backfila itself

PR #47, opened 2020-01-07
Configure Kotlin in Backfila's Moshi instance

swankjesse:


adrw:
Moshi needs the adaptor to correctly serialize/deserialize Kotlin data classes.

PR #51, opened 2020-01-13
Refactors how finding bounding ranges is selected. Adds strategies for Vitess.

mpawliszyn:


client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BoundingRangeStrategy.kt
```diff
@@ -0,0 +1,145 @@
...
+import app.cash.backfila.client.misk.InstanceProvider
+import app.cash.backfila.protos.clientservice.KeyRange
+import misk.hibernate.DbEntity
+import misk.hibernate.Keyspace
+import misk.hibernate.Session
+import misk.hibernate.Transacter
+import misk.hibernate.shards
+import misk.hibernate.transaction
+import okio.ByteString
+import org.glassfish.jersey.internal.guava.Ordering
```
shellderp: wrong import?

mpawliszyn: Yep good catch. Fixed

PR #58, opened 2020-01-29
Fix development app

shellderp:
By installing MiskWebModule

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
@@ -21,6 +21,7 @@ import misk.inject.KAbstractModule
 import misk.security.authz.AccessAnnotationEntry
 import misk.slack.SlackModule
 import misk.web.dashboard.AdminDashboardAccess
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
```diff
@@ -36,6 +37,7 @@ class BackfilaServiceModule(
     install(BackfilaWebActionsModule())
     install(BackfilaDashboardModule(environment))
     install(ServiceWebActionsModule())
+    install(AdminDashboardModule(environment))
```
shellderp: confused why this wasnt needed before.. is this installed in CashBackfilaService?

adrw: It's installed by Skim

shellderp: is it going to be broken if it's double installed?

shellderp: will merge, we can figure it out if it does break

PR #61, opened 2020-02-07
Add api dependency to misk hibernate

zhxnlai:
To make intellij happy

Screen Shot 2020-02-07 at 2 35 54 PM

### DIFF
--- a/client-misk/build.gradle
+++ b/client-misk/build.gradle
@@ -19,8 +19,8 @@ dependencies {
   api project(':client')
   testApi project(':client-misk-testing')
 
-  implementation(dep.misk)
-  implementation(dep.miskHibernate)
+  implementation dep.misk
+  api dep.miskHibernate
   testImplementation(dep.miskTesting)
   testImplementation(dep.miskHibernateTesting)
   testImplementation dep.junitEngine


client-misk/build.gradle
```diff
@@ -19,8 +19,8 @@ dependencies {
   api project(':client')
   testApi project(':client-misk-testing')
 
-  implementation(dep.misk)
-  implementation(dep.miskHibernate)
+  implementation dep.misk
+  api dep.miskHibernate
   testImplementation(dep.miskTesting)
```
adrw: Can you also remove the unnecessary brackets for other dep lines?

client-misk/build.gradle
```diff
@@ -19,8 +19,8 @@ dependencies {
   api project(':client')
   testApi project(':client-misk-testing')
 
-  implementation(dep.misk)
-  implementation(dep.miskHibernate)
+  implementation dep.misk
+  api dep.miskHibernate
```
mpawliszyn: Why am I not having this problem?

mpawliszyn:
> I don't think this is the right place to do this. We should do this in misk I think.

I'm wrong.

PR #62, opened 2020-02-08
Support nested path

zhxnlai:


adrw:
This seems like something where tests would be useful, can you add some?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfillOperator.kt
```diff
@@ -295,6 +297,15 @@ internal class BackfillOperator<E : DbEntity<E>, Pkey : Any> internal constructo
     return RunBatchResponse.Builder().build()
   }
 
+  private fun Backfill<*, *>.getPrimaryKeyPath(queryRoot: Root<*>): Path<Number> {
```
mpawliszyn: I think the place to fix this is in Misk? @swankjesse Thoughts?

mpawliszyn: @zhxnlai I'm not sure we want that other PR either?

zhxnlai: @mpawliszyn I think the fix should be in backfila because CriteriaBuilder is a JPA API: https://www.objectdb.com/java/jpa/query/criteria

swankjesse:
I like this PR as is. I think Misk would only do it if this was a Misk query API.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfillOperator.kt
```diff
@@ -102,8 +104,8 @@ internal class BackfillOperator<E : DbEntity<E>, Pkey : Any> internal constructo
       val minmax = queryFactory.dynamicQuery(backfill.entityClass)
           .dynamicUniqueResult(session) { criteriaBuilder, queryRoot ->
             criteriaBuilder.tuple(
-                criteriaBuilder.min(queryRoot.get(backfill.primaryKeyHibernateName())),
-                criteriaBuilder.max(queryRoot.get(backfill.primaryKeyHibernateName())))
+                criteriaBuilder.min(backfill.getPrimaryKeyPath(queryRoot)),
```
mpawliszyn: @swankjesse My only objection is that I can see these kinds of situations being somewhat common for performance reasons. Although maybe we just need a better way of specifying projections?

PR #66, opened 2020-02-12
Fix the TestingModule.

mpawliszyn:


client-misk/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
```diff
@@ -55,6 +55,6 @@ internal class ClientMiskTestingModule(
             url = "test.url", slack_channel = "#test"),
         backfillClasses
     ))
-    install(BackfilaClientModule())
+    install(EmbeddedBackfilaModule())
```
zhxnlai: An EmbeddedBackfilaModule is already added in L52

PR #67, opened 2020-02-13
Adds a way to specify default HttpClientEndpointConfig config for each connection type.

mpawliszyn:


service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClientProvider.kt
```diff
@@ -11,6 +15,14 @@ val DEFAULT_HTTP_CLIENT_ENDPOINT_CONFIG = HttpClientEndpointConfig(
     maxRequestsPerHost = 1000
 )
 
+class BackfilaDefaultEndpointConfigModule() : KAbstractModule() {
+  @Provides @Named(Connectors.HTTP)
+  fun httpClientEndpointConfigDefault() = DEFAULT_HTTP_CLIENT_ENDPOINT_CONFIG
+
+  @Provides @Named(Connectors.ENVOY)
```
swankjesse: Why isn't this in the envoy module?

service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClientProvider.kt
```diff
@@ -1,7 +1,11 @@
 package app.cash.backfila.client
 
+import app.cash.backfila.service.BackfilaConfig
+import com.google.inject.Provides
 import java.time.Duration
 import misk.client.HttpClientEndpointConfig
+import misk.inject.KAbstractModule
+import javax.inject.Named
```
shellderp: imports are out of order can you change your intellij config?

PR #71, opened 2020-02-18
Add javadoc to BackfillConfig.parameters and accessor function "parameter"

nurous:


client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfillConfig.kt
```diff
@@ -3,6 +3,15 @@ package app.cash.backfila.client.misk
 import okio.ByteString
 
 data class BackfillConfig(
+  /**
```
adrw: nit: Generally preferred to have KDocs be on the class instead of on the member/val

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfillConfig.kt
```diff
@@ -3,6 +3,15 @@ package app.cash.backfila.client.misk
...
+   * Populated from class level @Parameter annotations
+   * Use parameter function to get just the value (as a String)
+   * @see com.squareup.backfila.client.Parameter
+   */
   val parameters: Map<String, ByteString>,
   val dryRun: Boolean = true
-)
+) {
+  fun parameter(name: String): String? {
```
adrw: nit: `getParameterUtf8` or `getParameter` would be a bit more descriptive function name

PR #78, opened 2020-03-20
Set a title on backfila pages

shellderp:


shellderp:
> Probably will work 👍 If you want to edit/set other attributes like title, React Helmet is a great library for that.
> 
> https://github.com/nfl/react-helmet

cool could be useful to set title by page

PR #83, opened 2020-04-28
Bump Misk-Web and fix deprecated AdminDashboardModule

adrw:
0.1.26 fixes a dashboard bug where environment (staging vs production) would not show up for Backfila


adrw:
Environment labels are back 🎉 

Screen Shot 2020-04-28 at 11 01 38

PR #88, opened 2020-05-25
Add number of running backfills to services list

shellderp:


service/src/main/kotlin/app/cash/backfila/dashboard/GetServicesAction.kt
```diff
@@ -34,7 +43,16 @@ class GetServicesAction @Inject constructor(
       val services = queryFactory.newQuery<ServiceQuery>()
           .orderByName()
           .list(session)
-      services.map { it.registry_name }
+      val runningByService = queryFactory.newQuery<BackfillRunQuery>()
+          .state(BackfillState.RUNNING)
+          .list(session)
```
shellderp, line 48: no groupby in query factory, but this isn't too bad

mpawliszyn: Yeah this will be fine for now. Can optimize later when it is a problem.

PR #89, opened 2020-05-25
Add stack trace to runbatch response

shellderp:
So we can show it in UI / event log, I don't think we can get a useful trace from a 500


service/src/main/kotlin/app/cash/backfila/service/BatchAwaiter.kt
```diff
@@ -74,6 +80,10 @@ class BatchAwaiter(
           logger.info(e) { "Rpc failure when running batch for ${backfillRunner.logLabel()}" }
           backfillRunner.onRpcFailure()
 
+          if (e is RunBatchException) {
+            // TODO: write exception to event log
```
mpawliszyn, line 84: YES!!!

PR #91, opened 2020-06-04
Remove annotation based map

bezmax:
Preparatory refactor before extracting the endpoint serialization into Misk. 

Makes injection of connectors simpler, without use of additional annotation and multimap.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
@@ -1,10 +1,9 @@
 package app.cash.backfila.service
 
 import app.cash.backfila.api.ServiceWebActionsModule
-import app.cash.backfila.client.BackfilaClientServiceClientProvider
+import app.cash.backfila.client.ConnectorProvider
 import app.cash.backfila.client.Connectors
 import app.cash.backfila.client.EnvoyClientServiceClientProvider
-import app.cash.backfila.client.ForConnectors
 import app.cash.backfila.client.HttpClientServiceClientProvider
 import app.cash.backfila.dashboard.BackfilaDashboardModule
 import app.cash.backfila.dashboard.BackfilaWebActionsModule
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/app/cash/backfila/service/BackfilaServiceModule.kt
```diff
@@ -61,6 +53,15 @@ class BackfilaServiceModule(
     // something along the lines of requireBinding but works for multibindings.
   }
 
+  @Provides @Singleton
+  fun connectorProvider(
+    httpProvider: HttpClientServiceClientProvider,
+      envoyProvider: EnvoyClientServiceClientProvider
+  ) = ConnectorProvider(
```
shellderp, line 60: the loss of functionality here is that a service that pulls this in as a library can no longer add connector types

bezmax: Yes. The idea here is that:
1. Connectivity (including callbacks, and most of the code in this PR) will be extracted into Misk (separate PR that I'm working on).
2. The only way to add new connectors is to modify Misk code directly. We do not support pluggable connectors apart from what Misk supports, at least not right now.

I'm guessing the problem here is if Backfila is used with non-Misk based services, I'll need to clarify if we want to support this use-case.

shellderp: My concern is not non-misk but proprietary connectors. Maybe this isn't realistic as we haven't had to add proprietary connectors so far. Is there a bigger picture reason to pull this into misk?

bezmax: Right now we are trying to make it easier to integrate Backfila with other Misk services and Misk-supported infrastructure. Part of this work is to extract this connectivity code into Misk itself. This would allow building other services like Backfila (with registration of callbacks) easier, as the general workflow would look like this:

1. Ask Misk to give you your own callback blob. Misk collects the required environment information (envoy/non-envoy) and serializes that into proto blob.
2. Send this blob over the wire to Backfila/other service that works with callbacks.
3. Other service gives the blob again to Misk, where Misk is able to decode information in there back into an Http client.

For this flow to work, both client and server needs to be Misk-based. And at this moment, Misk is very opinionated about how services should communicate with eachother.

Maybe @swankjesse has thoughts on support of proprietary connectors.

shellderp: >For this flow to work, both client and server needs to be Misk-based.

don't you need to support square internal services that aren't misk based? at this point there's nothing about the "callback client" that makes it tied to misk

bezmax: For internal services we already kind-of reverse-engineer this, so this won't really change. Anyway, thanks for your input, we will think of how to make it more flexible.

PR #94, opened 2020-06-15
Migrating Environment to Deployment

mpawliszyn:
Might as well get this out of the way.


service/src/main/kotlin/app/cash/backfila/service/BackfilaDevelopmentService.kt
```diff
@@ -54,7 +61,7 @@ fun main(args: Array<String>) {
               slack = null
           )
       ),
-      AdminDashboardModule(Environment.DEVELOPMENT),
+      AdminDashboardModule(true),
```
shellderp, line 64: named arg would be nice

PR #96, opened 2020-06-17
Renaming Instance to Partition.

### LINKED ISSUE #50: Change Instance(s) to cursor(s)
I'm finding several people are getting confused on what an instance is. It is a term that is already pretty well overloaded.

I think a better name might be cursor.

mpawliszyn:
Closes #50 

I'm also doing a migration in the database. If you think this is too far LMK.


client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -30,16 +30,16 @@ message PrepareBackfillRequest {
   optional bool dry_run = 5;
 }
 message PrepareBackfillResponse {
-  message Instance {
-    optional string instance_name = 1;
+  message Partition {
+    optional string partition_name = 1;
```
shellderp, line 34: it's all proto encoded so this should be safe

service/src/main/kotlin/app/cash/backfila/client/HttpClientServiceClientProvider.kt
```diff
@@ -11,17 +14,14 @@ import okhttp3.Interceptor
...
-import javax.inject.Singleton
 
 @Singleton
 class HttpClientServiceClientProvider @Inject constructor(
-    @Named(HTTP) private val defaultHttpClientEndpointConfig: HttpClientEndpointConfig,
-    private val httpClientFactory: HttpClientFactory,
-    private val httpClientConfigUrlProvider: HttpClientConfigUrlProvider,
-    @HttpClientNetworkInterceptor private val networkInterceptors: List<Interceptor>,
-    private val moshi: Moshi
+  @Named(HTTP) private val defaultHttpClientEndpointConfig: HttpClientEndpointConfig,
```
shellderp, line 20: im confused where 2 space hanging indent is coming from

service/src/main/resources/migrations/v12__backfila.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE run_instances
```
shellderp, line 1: looks good

client/src/main/proto/app/cash/backfila/client_service.proto
mpawliszyn: Yeah I believe so.

service/src/main/kotlin/app/cash/backfila/client/HttpClientServiceClientProvider.kt
mpawliszyn: Yeah me too. I'm not seeing it in intellij

service/src/main/resources/migrations/v12__backfila.sql
mpawliszyn: Yeah and tests work. Should be safe back to 5.6.

client/src/main/proto/app/cash/backfila/client_service.proto
mpawliszyn: I think the only potential minor problem is that any of the old javascript pages might have trouble post a deploy. A simple refresh will fix it.

service/src/main/resources/migrations/v12__backfila.sql
shellderp: mysql 5.6? who still uses that? :)

PR #101, opened 2020-06-23
add parameters misk client

hdou731:


client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaStartupConfigurator.kt
```diff
@@ -35,8 +37,13 @@ internal class BackfilaStartupConfigurator @Inject internal constructor(
     val request = ConfigureServiceRequest.Builder()
         .backfills(
             backfills.values.map { backfillClass ->
+              // Create an instance of the Backfill so we can ask it what its parameters are. This
```
shellderp, line 40: the alternative would be annotations which you could retrieve without instantiating, has that been considered?

shellderp: also, as @mpawliszyn suggested, we could have a data class that contains parameters, eg
```
data class MyBackfillParams(
  val aString: String,
  val anInt: Int
)

@Parameters(MyBackfillParams::class)
class MyBackfill ...
```
that, or we add it as another generic type on the class. A way to avoid clutter for nonparameterized backfills could be to have `ParameterizedBackfill<Entity, Pkey, Parameters>` be the base class and another `Backfill<Entity, Pkey>` class that extends it and provides void parameters

I would treat each field as one parameter at the API level, so for that example you'd have two input fields, the `aString` and `anInt`, and if it's a complex object it can be json encoded

PR #102, opened 2020-06-24
Multibind backfills

shellderp:


client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaStartupConfigurator.kt
```diff
@@ -24,7 +24,7 @@ internal class BackfilaStartupConfigurator @Inject internal constructor(
   private val config: BackfilaClientConfig,
   private val backfilaClient: BackfilaClient,
   @ForBackfila private val moshi: Moshi,
-  @ForBackfila private val backfills: Map<String, KClass<out Backfill<*, *>>>
+  @ForBackfila private val backfills: MutableMap<String, KClass<out Backfill<*, *>>>
```
shellderp, line 27: has to be mutable map for mapbinding to work

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
```diff
@@ -44,5 +51,21 @@ class BackfilaModule(
...
 }
 
+private fun mapBinder(binder: Binder) = MapBinder.newMapBinder(
+    binder,
+    object : TypeLiteral<String>() {},
+    object : TypeLiteral<KClass<out Backfill<*, *>>>() {},
+    ForBackfila::class.java
+)
+
+inline fun <reified T : Backfill<*, *>> Binder.bindBackfill() {
```
shellderp, line 61: is there a smarter way to do this?

client-misk/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
```diff
@@ -48,10 +47,14 @@ internal class ClientMiskTestingModule(
...
-    ))
+    install(
+        BackfilaModule(
+            BackfilaClientConfig(
+                url = "test.url", slack_channel = "#test"
+            )
+        )
+    )
+
+    binder().bindBackfill<TestBackfill>()
```
shellderp, line 58: eg

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
mpawliszyn: I'm no expert but I think that's right.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaStartupConfigurator.kt
mpawliszyn: Do we want a point at which it is no longer mutable?

client-misk/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
mpawliszyn: Yeah this reified version looks good. 

We probably want to log what gets registered. I think this already happens on the backfila side though.

mpawliszyn, line 58: Is this the right place for this though?
Really I only want this to register for the SinglePartition use case.

We want to specify the backfills to register. Either by passing in the classes or adding a function that does the binding.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
mpawliszyn: Maybe @swankjesse has a suggestion but I believe this is correct.

client-misk/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
shellderp: I think it is the right place and we would just have multiple backfill classes for different tests, wdyt?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaStartupConfigurator.kt
shellderp: the two cases we have are used in very small scopes so I'm not worried

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
```diff
@@ -4,37 +4,44 @@ import app.cash.backfila.client.misk.client.BackfilaClientConfig
...
-    for (backfill in backfills) {
-      map[backfill.jvmName] = backfill
+    if (backfills == null) {
+      mapBinder(binder())
+    } else {
+      val map = mutableMapOf<String, KClass<out Backfill<*, *>>>()
+      for (backfill in backfills) {
+        map[backfill.jvmName] = backfill
+      }
+      bind(object : TypeLiteral<MutableMap<String, KClass<out Backfill<*, *>>>>() {})
```
swankjesse, line 41: why not multibind them here? Then you can gradually transition from non-multibindings to multibindings

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
swankjesse: I prefer a slightly different syntax where we let the application itself do the multibind() call.

Here’s how we document it in StaticResourceEntry:
```
/**
 * This data class is used with Guice multibindings. Register instances by calling `multibind()`
 * in a `KAbstractModule`:
 *
 * ```
 * multibind<StaticResourceEntry>().toInstance(StaticResourceEntry(...))
 * ```
 */
class StaticResourceEntry(
  url_path_prefix: String = "/",
  private val resourcePath: String
) : ValidWebEntry(url_path_prefix = url_path_prefix) {
 ...
}
```

swankjesse: Or perhaps what we do for Web Actions?
```
    install(WebActionModule.create<InternalErrorAction>())
```

swankjesse: (this way you don’t need to subclass a special module class, which I find a bit awkward)

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
shellderp: doh, fixed

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
shellderp: thanks, that's way better

client-misk/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
shellderp: done

client-misk/src/test/kotlin/app/cash/backfila/client/misk/hibernate/SinglePartitionHibernateBackfillTest.kt
```diff
@@ -1,64 +1,27 @@
...
 import misk.hibernate.Id
-import misk.hibernate.Query
 import misk.hibernate.Session
 import misk.hibernate.Transacter
 import okhttp3.internal.toImmutableList
-import okio.ByteString
 import okio.ByteString.Companion.encodeUtf8
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
+import javax.inject.Inject
```
shellderp, line 17: if someone knows how to fix intellij doing this..

PR #115, opened 2020-07-31
Migrate to using HttpClientsConfig.

mpawliszyn:
This opens the door to replacing the connectorExtraData with just a url field I believe.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClientProvider.kt
+++ b/service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClientProvider.kt
@@ -1,25 +1,28 @@
 package app.cash.backfila.client
 
 import com.google.inject.Provides
+import misk.client.HttpClientConfig
 import java.time.Duration
 import javax.inject.Named
 import misk.client.HttpClientEndpointConfig
+import misk.client.HttpClientsConfig
 import misk.inject.KAbstractModule
 
-val DEFAULT_HTTP_CLIENT_ENDPOINT_CONFIG = HttpClientEndpointConfig(
-    // Allow RunBatch requests to take a long time.
-    readTimeout = Duration.ofSeconds(30L),
-    // Use a large number, we limit parallel calls ourselves.
-    maxRequests = 1000,
-    maxRequestsPerHost = 1000
+val DEFAULT_HTTP_CLIENTS_CONFIG = HttpClientsConfig(
+  hostConfigs = linkedMapOf(
+    ".*" to HttpClientConfig(
+      // Allow RunBatch requests to take a long time.
+      readTimeout = Duration.ofSeconds(30L),
+      // Use a large number, we limit parallel calls ourselves.
+      maxRequests = 1000,
+      maxRequestsPerHost = 1000
+    )
+  )
 )
 
 class BackfilaDefaultEndpointConfigModule() : KAbstractModule() {
-  @Provides @Named(Connectors.HTTP)
-  fun httpClientEndpointConfigDefault() = DEFAULT_HTTP_CLIENT_ENDPOINT_CONFIG
-
-  @Provides @Named(Connectors.ENVOY)
-  fun envoyClientEndpointConfigDefault() = DEFAULT_HTTP_CLIENT_ENDPOINT_CONFIG
+  @Provides
+  fun httpClientsConfigDefault() = DEFAULT_HTTP_CLIENTS_CONFIG
 }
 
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/app/cash/backfila/client/EnvoyClientServiceClientProvider.kt
```diff
@@ -2,23 +2,25 @@ package app.cash.backfila.client
...
 import retrofit2.adapter.guava.GuavaCallAdapterFactory
 import retrofit2.converter.wire.WireConverterFactory
 
 @Singleton
 class EnvoyClientServiceClientProvider @Inject constructor(
-  @Named(ENVOY) private val defaultHttpClientEndpointConfig: HttpClientEndpointConfig,
+  private val httpClientsConfig: HttpClientsConfig,
   private val httpClientFactory: HttpClientFactory,
-  private val httpClientConfigUrlProvider: HttpClientConfigUrlProvider,
+  private val envoyClientEndpointProvider: EnvoyClientEndpointProvider,
```
mpawliszyn, line 23: @shellderp what do you think of replacing connectorExtraData with just a url field?

shellderp: sounds trivial for http

```
data class HttpConnectorData(val url: String)

data class EnvoyConnectorData(val clusterType: String)
```

for envoy would `clusterType` be the "url"?

shellderp: seems reasonable, ultimately it's just a callback url

service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClientProvider.kt
```diff
@@ -1,25 +1,28 @@
...
 
-val DEFAULT_HTTP_CLIENT_ENDPOINT_CONFIG = HttpClientEndpointConfig(
-    // Allow RunBatch requests to take a long time.
-    readTimeout = Duration.ofSeconds(30L),
-    // Use a large number, we limit parallel calls ourselves.
-    maxRequests = 1000,
-    maxRequestsPerHost = 1000
+val DEFAULT_HTTP_CLIENTS_CONFIG = HttpClientsConfig(
+  hostConfigs = linkedMapOf(
+    ".*" to HttpClientConfig(
```
shellderp, line 13: huh, is regex in the key a new feature?

mpawliszyn: The idea is that backfila is no longer in the middle. Its how you define what urls get what config.

service/src/main/kotlin/app/cash/backfila/client/EnvoyClientServiceClientProvider.kt
mpawliszyn: Well I was thinking of just replacing the data type completely. Although now I'm thinking that is just something to do when we run into the next client type?

Unfortunately the retrofit clients are still a bit different between the envoy and http clients. Unless you think that should be specified too.

service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClientProvider.kt
bezmax: Yeah, we tried to make routing of requests abstracted away. For this to work, I changed the HttpClientsConfig to be URL based. This way, Skim can define default routing settings (like `*.gns.square` always going through the abstract socket), and clients can just provide an URL and delegate the correct routing of this URL to Misk/Skim/OS.

service/src/main/kotlin/app/cash/backfila/client/EnvoyClientServiceClientProvider.kt
bezmax: @mpawliszyn Could you clarify why retrofit clients are different? From what I've seen, to build a retrofit client you still go to HttpClientsConfig to ask it for config, and they you build the client from that. Therefore, same approach still applies - as long as you know the URL you can build the client. Or am I missing something?

shellderp: yea I meant replace data structure with url. exactly what does the url look like for the envoy client though? I suppose that's not my problem :p

PR #116, opened 2020-08-04
Add management client

shellderp:


client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaManagementClient.kt
```diff
@@ -0,0 +1,18 @@
+package app.cash.backfila.client.misk
+
+import okio.ByteString
+
+interface BackfilaManagementClient {
+  fun <B : Backfill<*, *, P>, P : Any> createAndStart(
+    backfillClass: Class<B>,
```
mpawliszyn, line 7: Do we want any defaults here?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/RealBackfilaManagementClient.kt
```diff
@@ -0,0 +1,47 @@
+package app.cash.backfila.client.misk
+
+import app.cash.backfila.client.misk.internal.BackfilaClient
+import app.cash.backfila.client.misk.internal.parametersToBytes
+import app.cash.backfila.protos.service.CreateAndStartBackfillRequest
+import app.cash.backfila.protos.service.CreateBackfillRequest
+import okio.ByteString
+import javax.inject.Inject
+
+internal class RealBackfilaManagementClient @Inject internal constructor(
```
mpawliszyn, line 10: Don't we want a fake version of this that we can inject??

mpawliszyn: NVM I looked at the code. The existing embedded bypasses this I guess.

PR #120, opened 2020-08-10
Create build action to upload snapshots

shellderp:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,11 +1,9 @@
 name: build
 
 on:
-  push
-
-#env:
-#  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
-
+  push:
+    branches:
+      - master
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,7 @@
 name: build
 
 on:
-  push:
-    branches:
-      - master
+  push
 
 #env:
 #  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: build
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
+name: build
+
+on:
+  push:
+    branches:
+      - master
```
mpawliszyn, line 6: So it releases on every master push?

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+          npm install -g @misk/cli
+          miskweb ci-build -e
+      - name: Cache Gradle packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Set gradle version
+        run: echo "version=String.valueOf(project.version).replace(\"-SNAPSHOT\", \"\") + '-$(date +%Y%m%d)-$(git rev-parse --short HEAD)-SNAPSHOT'" > hooks.gradle
```
mpawliszyn, line 32: But publishes with a date and rev number? I guess this is ok? Do other packages do this?

.github/workflows/build.yml
shellderp: yea

.github/workflows/build.yml
shellderp: SNAPSHOTs get overwritten when you push, so this fixes that problem by having a unique version for each one. Not exactly following conventions but maven conventions suck :p

shellderp:
> Just a couple questions on how happy Sonatype Nexus will be with this

shrug, we don't have that many master pushes anyway. Wire does a snapshot release on every master and seems fine. They don't do the date version thing but I don't see how it's different, the same # of artifacts are stored

shellderp:
I need to look into why there's no actual jar pushed to maven.. wire seems to have the same problem right now. Only sources and docs jars are pushed

shellderp:
> I need to look into why there's no actual jar pushed to maven.. wire seems to have the same problem right now. Only sources and docs jars are pushed

nevermind it works fine

.github/workflows/build.yml
mpawliszyn: SG

PR #127, opened 2020-09-09
Don't install the obsolete JUnit Gradle plugin

swankjesse:
Use the built-in one instead.


mpawliszyn (approved):
🎉🎉🎉 Fixes always running all the tests. 🎉🎉🎉

mpawliszyn:
@swankjesse I think you forgot to include the removal from the dependencies.gradle file though?

mpawliszyn:
That should fix it.

PR #131, opened 2020-09-14
Make the backfila API work even when degraded.

mpawliszyn:
Backfila will inherently stress out your service so regularly scheduled
concurrency limiters tend to overreact. If problematic limits are hit
then an error is generally produced and backfila will quickly back off.

### DIFF
--- a/client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaWebApiActions.kt
+++ b/client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaWebApiActions.kt
@@ -10,6 +10,7 @@ import app.cash.backfila.protos.clientservice.RunBatchResponse
 import javax.inject.Inject
 import misk.logging.getLogger
 import misk.security.authz.Authenticated
+import misk.web.AvailableWhenDegraded
 import misk.web.Post
 import misk.web.RequestBody
 import misk.web.RequestContentType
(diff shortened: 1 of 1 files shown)


client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaWebApiActions.kt
```diff
@@ -27,6 +28,7 @@ internal class PrepareBackfillAction @Inject constructor(
   @ResponseContentType(MediaTypes.APPLICATION_PROTOBUF)
   @Authenticated(services = ["backfila"])
   @LogRequestResponse(bodySampling = 1.0, errorBodySampling = 1.0)
+  @AvailableWhenDegraded
```
mightyguava, line 31: I thought this annotation is supposed to go on the server-side

mpawliszyn: This is the API that gets created by the client that backfila uses to call back in.

mightyguava: nvm, this is the server...

shellderp (approved):
I thought we already did this lol. Definitely makes sense

swankjesse (approved):

PR #141, opened 2020-10-15
A bit more refactoring to separate the hibernate client implementation.

mpawliszyn:
This allows the client implemetations to use any kind of class they
wish and separates the installing of the client implementation from the
base communication to/from backfila.

There is still a todo to separate this in the
BackfilaStartupConfigurator but that will be in a follow-up.

### DIFF
--- a/client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
+++ b/client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
@@ -2,83 +2,57 @@ package app.cash.backfila.client.misk
 
 import app.cash.backfila.client.misk.client.BackfilaClientConfig
 import app.cash.backfila.client.misk.client.BackfilaClientLoggingSetupProvider
+import app.cash.backfila.client.misk.client.BackfilaClientModule
 import app.cash.backfila.client.misk.client.BackfilaClientNoLoggingSetupProvider
+import app.cash.backfila.client.misk.embedded.EmbeddedBackfilaModule
 import app.cash.backfila.client.misk.internal.BackfilaClient
 import app.cash.backfila.client.misk.internal.BackfilaStartupConfigurator
-import app.cash.backfila.client.misk.internal.BackfillOperator
-import app.cash.backfila.client.misk.internal.HibernateBackfillOperator
 import app.cash.backfila.client.misk.internal.RealBackfilaClient
-import com.google.inject.Binder
 import com.google.inject.BindingAnnotation
 import com.google.inject.Provides
-import com.google.inject.TypeLiteral
-import com.google.inject.multibindings.MapBinder
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import javax.inject.Singleton
 import kotlin.reflect.KClass
-import kotlin.reflect.jvm.jvmName
 import misk.ServiceModule
 import misk.inject.KAbstractModule
 
 /**
- * Backfila-using applications install this and either [EmbeddedBackfilaModule] (testing and
- * development) or [BackfilaClientModule] (staging and production).
+ * Backfila-using applications install this, one or more specific client implementation modules such
+ * as [HibernateBackfilaClientModule] and either [EmbeddedBackfilaModule] (testing and development)
+ * or [BackfilaClientModule] (staging and production).
  */
 class BackfilaModule(
   private val config: BackfilaClientConfig,
-  @Deprecated(message = "Multibind backfills instead")
+  @Deprecated(message = "Multibind backfills using 'HibernateBackfillInstallModule.create' instead")
   private val backfills: List<KClass<out Backfill<*, *, *>>>? = null,
   private val loggingSetupProvider: KClass<out BackfilaClientLoggingSetupProvider> =
       BackfilaClientNoLoggingSetupProvider::class
 ) : KAbstractModule() {
   override fun configure() {
     bind<BackfilaClientConfig>().toInstance(config)
-    multibind<BackfillOperator.Backend>().to<HibernateBackfillOperator.HibernateBackend>()
 
     bind<BackfilaClient>().to<RealBackfilaClient>()
     bind<BackfilaManagementClient>().to<RealBackfilaManagementClient>()
     bind<BackfilaClientLoggingSetupProvider>().to(loggingSetupProvider.java)
 
-    mapBinder(binder())
+    // For backwards compatibility for now we install the Hibernate Backfila Client implementation
+    // and support the old backfills parameter.
+    install(HibernateBackfilaClientModule())
 
     if (backfills != null) {
       for (backfill in backfills) {
-        install(BackfillInstallModule.create(backfill))
+        install(HibernateBackfillInstallModule.create(backfill))
       }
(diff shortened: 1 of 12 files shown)


mpawliszyn:
I've now added the follow up commit that handles the startup configurator.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
```diff
@@ -2,83 +2,57 @@ package app.cash.backfila.client.misk
...
- * Backfila-using applications install this and either [EmbeddedBackfilaModule] (testing and
- * development) or [BackfilaClientModule] (staging and production).
+ * Backfila-using applications install this, one or more specific client implementation modules such
+ * as [HibernateBackfilaClientModule] and either [EmbeddedBackfilaModule] (testing and development)
+ * or [BackfilaClientModule] (staging and production).
  */
 class BackfilaModule(
   private val config: BackfilaClientConfig,
-  @Deprecated(message = "Multibind backfills instead")
+  @Deprecated(message = "Multibind backfills using 'HibernateBackfillInstallModule.create' instead")
```
shellderp, line 27: have you converted off of this? can remove if so

client-misk/src/main/kotlin/app/cash/backfila/client/misk/HibernateBackfilaClientModule.kt
```diff
@@ -0,0 +1,42 @@
...
+
+class HibernateBackfilaClientModule() : KAbstractModule() {
+  override fun configure() {
+    multibind<BackfillOperator.Backend>().to<HibernateBackfillOperator.HibernateBackend>()
+
+    mapBinder(binder()) // Default the mapbinder to having no Backfills registered.
+  }
+}
+
+class HibernateBackfillInstallModule<T : Backfill<*, *, *>> private constructor(
```
shellderp, line 20: I guess I don't agree with this degree of separation - are  we going to duplicate this for each base type of backfill? could we not have one Backfill interface that they all implement?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/embedded/Backfila.kt
```diff
@@ -11,14 +10,14 @@ import okio.ByteString
  * the Backfila dashboard UI.
  */
 interface Backfila {
-  fun <Type : Backfill<*, *, *>> createDryRun(
+  fun <Type : Any> createDryRun(
```
shellderp, line 13: same strangeness here.. why not have a top Backfill interface that DynamoBackfill, HibernateBackfill etc implement?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/HibernateBackfillOperator.kt
```diff
@@ -342,6 +349,25 @@ internal class HibernateBackfillOperator<E : DbEntity<E>, Pkey : Any, Param : An
...
 
+    override fun backfills(): Set<BackfillOperator.BackfillRegistration> {
+      return backfills.map { BackfillOperator.BackfillRegistration(
+          name = it.key,
+          description = (it.value.annotations.find { it is Description } as? Description)?.text,
+          parametersClass = parametersClass(it.value as KClass<Backfill<*, *, Any>>)
+      ) }.toSet()
+    }
+
+    private fun <T : Any> parametersClass(backfillClass: KClass<out Backfill<*, *, T>>): KClass<T> {
```
shellderp, line 360: wouldn't this be shared across all types of backfills?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaModule.kt
mpawliszyn: Not entirely yet. Maybe I'll just do it. But I'll remove it in a later pr.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/embedded/Backfila.kt
mpawliszyn: So they can do radically different things. Have different base classes.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/HibernateBackfilaClientModule.kt
mpawliszyn: I think this is a reasonable thing to duplicate since the backend might have some other way of specifying the backfills. It allows them to diverge a bit more so we avoid locking them all together. Makes it a has-a rather than an is-a.

You could have a backend that generates the backfills.

For instance we may for dynamo we might always want to add certain parameters. This allows us to do this in many ways rather than being forced to have a super Parameters type.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/HibernateBackfillOperator.kt
mpawliszyn: No. Dynamo only has two Parameterized types. So this is different enough that it deserves some copy and paste.

I think it is the job of each implementation to specify how to calculate the Parameterized type.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/BackfilaStartupConfigurator.kt
```diff
@@ -26,23 +22,34 @@ internal class BackfilaStartupConfigurator @Inject internal constructor(
...
+
+    // Send the Backfila registration request
     val request = ConfigureServiceRequest.Builder()
         .backfills(
-            backfills.values.map { backfillClass ->
-              val description = (backfillClass.annotations.find { it is Description } as? Description)?.text
+            registrations.map { registration ->
               @Suppress("UNCHECKED_CAST")
-              val parameters = backfilaParametersForBackfill(backfillClass as KClass<Backfill<*, *, Any>>)
+              val parameters = backfilaParametersFromClass(registration.parametersClass)
```
mpawliszyn, line 49: I actually wonder @shellderp whether the implementations should be in-charge of providing the whole parameters blob rather than just the class

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/HibernateBackfillOperator.kt
mpawliszyn: Here is my current Dynamo version:
https://github.com/cashapp/backfila/blob/mikepaw.rough-dynamo/client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/DynamoDbBackfillOperator.kt#L213

client-misk/src/main/kotlin/app/cash/backfila/client/misk/HibernateBackfilaClientModule.kt
mpawliszyn: I think the implementations will all require massive changes to migrate a backfill between them. I think this is ok and actually what we want customers to do.

swankjesse:
Reviewed interactively.

swankjesse (approved):

PR #142, opened 2020-10-19
Refactoring the classes into packages that can be moved the better modules.

mpawliszyn:
This will break existing clients. I will follow up with another change
to break out the modules only.

This also creates a base Backfill type.

Please do not merge this before I have the follow up that moves the modules around.


mpawliszyn:
I've added the separate commit for adding the client-misk-hibernate module. Feel free to look at them separately.

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
```diff
@@ -51,19 +52,17 @@ internal class ClientMiskTestingModule(
...
-    )
+    ))
 
     bind(BackfilaClientLoggingSetupProvider::class.java)
         .to(BackfilaClientNoLoggingSetupProvider::class.java)
 
-    install(HibernateBackfillInstallModule.create<SinglePartitionHibernateTestBackfill>())
-    install(HibernateBackfillInstallModule.create<ChickenToBeefBackfill>())
-    install(HibernateBackfillInstallModule.create<RecordNoParametersConfigValuesBackfill>())
+    install(create<SinglePartitionHibernateTestBackfill>())
```
shellderp, line 64: not sure I like the readability of this

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaManagementClient.kt
```diff
@@ -3,11 +3,11 @@ package app.cash.backfila.client.misk
 import okio.ByteString
 
 interface BackfilaManagementClient {
-  fun <B : Backfill<*, *, P>, P : Any> createAndStart(
+  fun <B : Backfill> createAndStart(
     backfillClass: Class<B>,
     dry_run: Boolean,
     num_threads: Int? = null,
-    parameters: P? = null,
+    parameters: Any? = null,
```
shellderp, line 10: maybe we can eventually bring this back by having a Parameters type on the Backfill interface. makes sense given the API has the parameters built in

client-misk/src/main/kotlin/app/cash/backfila/client/misk/spi/BackfillBackend.kt
```diff
@@ -0,0 +1,10 @@
+package app.cash.backfila.client.misk.spi
+
+/**
+ * Service provider interface for backends like Hibernate and DynamoDb. Backends construct operators
+ * that actually run the backfill.
+ */
+interface BackfillBackend {
```
shellderp, line 7: this interface is great

shellderp (approved):

swankjesse (approved):

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
mpawliszyn: Which part? The install of the create or the fact that it's reified?

client-misk/src/main/kotlin/app/cash/backfila/client/misk/BackfilaManagementClient.kt
mpawliszyn: I guess that is the question. I do think that it can make sense to have parameters be part of every Backfill.

It means that certain Backfill types may have a base type of Parameters object. I do think that is ok. One drawback is that you won't be able to use some externally defined class directly. You would need something to wrap it. Which is probably not a huge deal TBH.

I also think that I want the test functions to continue to be able to take arbitrary objects.

client-misk/src/main/kotlin/app/cash/backfila/client/misk/spi/BackfillBackend.kt
mpawliszyn: Glad you like it! Kudos goes to @swankjesse too.

mpawliszyn:
I'm going to merge this in and do any comments as a follow up.

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/ClientMiskTestingModule.kt
shellderp: I preferred it without the static import, reading `install(create<>())` is a little odd but I guess it's totally fine just in this test module

PR #147, opened 2020-10-28
Creates a basic DynamoDbBackend that can mutate rows.

mpawliszyn:


client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/DynamoDbBackfill.kt
```diff
@@ -0,0 +1,39 @@
+package app.cash.backfila.client.misk.dynamodb
+
+import app.cash.backfila.client.misk.Backfill
+import app.cash.backfila.client.misk.BackfillConfig
+
+abstract class DynamoDbBackfill<I : Any, P : Any> : Backfill {
+
+  abstract val itemType: Class<I>
```
shellderp, line 8: couldn't you use the same trick to get this as in the hibernate one?

shellderp (approved):
I don't know much about dynamo but it seems reasonable

client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/internal/DynamoDbBackfillOperator.kt
```diff
@@ -0,0 +1,93 @@
...
+        totalSegments = count
+        limit = 4
+        if (lastEvaluatedKey != null) {
+          exclusiveStartKey = lastEvaluatedKey
+        }
+      }
+      val result = dynamoDb.scanPage(backfill.itemType, scanRequest)
+      backfill.runBatch(result.results, config)
+      lastEvaluatedKey = result.lastEvaluatedKey
+    } while (lastEvaluatedKey != null)
```
shellderp, line 88: why do you need to loop here? A batch can span multiple pages? how long would each batch take?

client-misk-dynamodb/README.md
```diff
@@ -0,0 +1,3 @@
+# client-misk-dynamodb
+
+Backfila Misk client implementation to backfill a DynamoDb datastore.
```
swankjesse, line 3: supernit: DynamoDB is the spelling with an uppercase B. We use DynamoDb in code because its an acronym

client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/DynamoDbBackfill.kt
swankjesse: good idea

swankjesse (approved):

client-misk-dynamodb/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=client-misk-dynamodb
+POM_NAME=client-misk-dynamodb
+POM_DESCRIPTION=Misk DynamoDb client library for backfila
```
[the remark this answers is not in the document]
mpawliszyn: Well it relies on the misk-client stuff. But otherwise yeah. I'm thinking it might make sense to eventually extract the misk part out. But for now it is misk specific.

client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/internal/DynamoDbBackfillOperator.kt
mpawliszyn: So this is where this is a basic version of the backfill. Yes a batch can span multiple pages. With this implementation it could yeah take a really long time.

We have three improvements:
1. During the prepare we are going to try and guess the number of segments to break up the backfill into a bit better. 
2. I'm going to add a way for run batch to essentially partially run, return with "call me again" information that backfila will just call again. Which would complete the next n-items. 
3. This is more risky but we will have an option of establishing batches that are parts of a segment. The contract with DyanamoDB indicates that it should work but since it isn't explicit it's a bit risky. This will use marker row information. If the marker row gets deleted then the backfill will fail. Resuming the backfill will work though.

mpawliszyn: If it wan't clear these will be coming in later PRs.

client-misk-dynamodb/README.md
mpawliszyn: Yep done. Here and in comments it is DynamoDB. Where it is part of code it is DynamoDb or dynamoDb.

client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/DynamoDbBackfill.kt
mpawliszyn: Yeah of course. Done

PR #148, opened 2020-11-04
Refactors the Runner and Awaiter states.

mpawliszyn:
I want to add a new reason for retrying a Batch. This refactors the code
a bit to accomplish this. In this PR we can also talk about any wider
refactor we might want to do.

I want to add a bookmark bytestring that when returned means retry the runBatch but include this string as well. This will allow for uneven batches to partially process and indicate to future selves where to pick up from. We could potentially use this  in hibernate too if we had a target or maximum latency like we do for calculating batches.

I looked into moving the global backoff in some kind of wrapper from the BackfillRunner instead of using the client directly but I think that ended up complicating things more.

Or maybe we will come up with a better wider refactor based on the comments.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
@@ -7,7 +7,6 @@ import com.google.common.base.Stopwatch
 import java.time.Duration
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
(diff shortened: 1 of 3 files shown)


swankjesse (approved):

service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
```diff
@@ -91,11 +92,7 @@ class BatchRunner(
         // update cursors.
         async { RunBatchResponse.Builder().build() }
```
shellderp, line 93: I think this part could also be moved into the new method

shellderp: nevermind, it will never need a retry :p

PR #149, opened 2020-11-06
Provides a partial batch bookmark so long batches can return early but retain progress.

mpawliszyn:
When enabled in a client this allows an underlying batch to return early
when it is at risk of timing out. Data can be placed in the bookmark on
how to continue the batch. The resulting repeat call contains this data.

### DIFF
--- a/client/src/main/proto/app/cash/backfila/client_service.proto
+++ b/client/src/main/proto/app/cash/backfila/client_service.proto
@@ -105,7 +105,16 @@ message RunBatchResponse {
 
   optional PipelinedData pipelined_data = 2;
 
+  // Stacktrace that will be surfaced to the user.
+  // If this is provided this batch request is assumed to have failed.
   optional string exception_stack_trace = 3;
+
+  // Provided data from the backfila client telling backfila what part of the batch was not
+  // completed yet. Backfila will end up resuming with a new request containing either the remaining
+  // range or the original range.
+  // This is used by Backfila clients to relay what progress was completed in situations like
+  // time consuming batches.
+  optional KeyRange remaining_batch_range = 8;
 }
 
 // Pipelined data can be any format as long as both the consumer and provider agree on it.
--- a/service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchAwaiter.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchAwaiter.kt
@@ -1,5 +1,6 @@
 package app.cash.backfila.service.runner.statemachine
 
+import app.cash.backfila.protos.clientservice.GetNextBatchRangeResponse
 import app.cash.backfila.protos.clientservice.RunBatchResponse
 import app.cash.backfila.service.persistence.BackfillState
 import app.cash.backfila.service.persistence.DbEventLog
@@ -29,7 +30,7 @@ class BatchAwaiter(
   ) = scope.launch {
     logger.info { "BatchAwaiter started ${backfillRunner.logLabel()}" }
     main@ while (true) {
-      var (batch, runBatchRpc, startedAt) = try {
+      val (initialBatch, initialRunBatchRpc, batchStartedAt) = try {
         receiveChannel.receive()
       } catch (e: CancellationException) {
         logger.info(e) { "BatchAwaiter job cancelled ${backfillRunner.logLabel()}" }
@@ -40,10 +41,12 @@ class BatchAwaiter(
         break
       }
 
+      var remainingBatch = initialBatch
+      var runBatchRpc = initialRunBatchRpc
+      var callStartedAt = batchStartedAt
       // Repeat this batch until it succeeds.
       retry@ while (true) {
-        var retry: Deferred<RunBatchResponse>? = null
-        try {
+        val nextRequest: Deferred<RunBatchResponse> = try {
           val response: RunBatchResponse = runBatchRpc.await()
 
           if (response.exception_stack_trace != null) {
@@ -57,6 +60,37 @@ class BatchAwaiter(
               backfillRunner.runBatchBackoff.addMillis(backfillRunner.metadata.extraSleepMs)
             }
           }
+
+          if (response.remaining_batch_range != null) {
+            // We have a remaining_batch_range, continue the batch.
+            remainingBatch = initialBatch.newBuilder()
+                .batch_range(response.remaining_batch_range)
+                .build()
+            backoffAndSendRunBatchAsync(remainingBatch) {
+              "${backfillRunner.logLabel()} continuing remaining range " +
+                  "${response.remaining_batch_range} of for partially completed batch $initialBatch"
+            }
+          } else {
+            // We got a 200 response on the batch with no error or partial bookmark. This batch is
+            // done.
+            logger.info { "Runbatch finished for ${backfillRunner.logLabel()} $initialBatch" }
+
+            backfillRunner.onRpcSuccess()
+
+            matchingRateCounter.add(initialBatch.matching_record_count)
+            scannedRateCounter.add(initialBatch.scanned_record_count)
+            // Track our progress in DB for when another runner takes over.
+            // TODO update this less often, probably in the lease updater task
+            backfillRunner.factory.transacter.transaction { session ->
+              val dbRunPartition = session.load(backfillRunner.partitionId)
+              dbRunPartition.pkey_cursor = initialBatch.batch_range.end
+              dbRunPartition.backfilled_scanned_record_count += initialBatch.scanned_record_count
+              dbRunPartition.backfilled_matching_record_count += initialBatch.matching_record_count
+              dbRunPartition.scanned_records_per_minute = scannedRateCounter.projectedRate()
+              dbRunPartition.matching_records_per_minute = matchingRateCounter.projectedRate()
+            }
+            break@retry
+          }
         } catch (e: CancellationException) {
           logger.info(e) { "BatchAwaiter job cancelled ${backfillRunner.logLabel()}" }
           break@main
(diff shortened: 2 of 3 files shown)


service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchAwaiter.kt
```diff
@@ -42,8 +44,7 @@ class BatchAwaiter(
 
       // Repeat this batch until it succeeds.
       retry@ while (true) {
-        var retry: Deferred<RunBatchResponse>? = null
-        try {
+        val nextRequest: Deferred<RunBatchResponse> = try {
```
mpawliszyn, line 47: I love kotlin

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -589,6 +589,147 @@ class BackfillRunnerTest {
...
+
+        // complete the two stuck requests currently being processed
+        fakeBackfilaClientServiceClient.runBatchResponses.send(
+            Result.success(RunBatchResponse.Builder().build()))
+        delay(1) // allow the other coroutine scope to complete TODO: yield to right scope
+        assertThat(getSinglePartitionCursor(runner)).isEqualTo(firstBatchCursor) // cursor is still unchanged
+        fakeBackfilaClientServiceClient.runBatchResponses.send(
+            Result.success(RunBatchResponse.Builder().build()))
+        delay(1) // allow the other coroutine scope to complete TODO: yield to right scope
+        val cursor = getSinglePartitionCursor(runner)
```
mpawliszyn, line 709: I'm open to suggestions on how to make the cursor checking better.

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -116,7 +116,7 @@ class BackfillRunnerTest {
       try {
         // We should only get numthreads=1 calls in parallel, then it must wait for more room.
         assertThat(fakeBackfilaClientServiceClient.runBatchRequests.receive()).isNotNull()
-        yield()
+        delay(1) // allow the other coroutine scope to complete TODO: yield to right scope
```
mpawliszyn, line 119: yield is on the wrong coroutine scope. I'll refactor this later. 

I'm wondering if the delay is better than poking the scope though to the tests.

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -589,6 +589,153 @@ class BackfillRunnerTest {
...
+
+        waitForOtherCoroutines()
+        assertThat(fakeBackfilaClientServiceClient.runBatchRequests.poll()).isNull()
+      } finally {
+        runner.stop()
+      }
+    }
+  }
+
+  private suspend fun waitForOtherCoroutines() {
```
mpawliszyn, line 726: This is slightly better than having that delay everywhere. Maybe this is sufficient?

shellderp: seems fine

client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -98,14 +98,24 @@ message RunBatchRequest {
...
+  optional bytes partial_batch_bookmark = 8;
 }
 message RunBatchResponse {
   // How much to sleep before starting new batches. For example can be set to replication lag.
   optional uint64 backoff_ms = 1;
 
   optional PipelinedData pipelined_data = 2;
 
+  // Stacktrace that will be surfaced to the user.
+  // If this is provided this batch request is assumed to have failed.
```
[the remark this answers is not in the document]
mpawliszyn: welcome

client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -98,14 +98,24 @@ message RunBatchRequest {
...
   optional string exception_stack_trace = 3;
+
+  // Provided data from the backfila client telling the next invocation how to continue.
+  // Allows the client to return even if a batch is not done.
+  // If this is provided this batch request is considered incomplete.
+  optional bytes partial_batch_bookmark = 8;
```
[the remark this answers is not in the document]
mpawliszyn: We had hack week last week which is why it is taking a while to get back to this.  Correct me if I'm wrong but I think you are suggesting:

```
  optional KeyRange remaining_range = 8;
}
```

It would save us from having this in the Request. I think I really like it?

So in the Dynamo case there will be extra data in this case and I wonder if it makes sense to store this data with the KeyRange data.

It might just make sense since this extra data is giving us information about what is left/done.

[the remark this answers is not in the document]
shellderp: yea that's what I'm thinking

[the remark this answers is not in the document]
mpawliszyn: Finally got this done. That test took a while. Turns out I needed to delay a few times to make sure all the other coroutines were done...

client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -95,17 +95,23 @@ message RunBatchRequest {
...
   optional uint64 backoff_ms = 1;
 
   optional PipelinedData pipelined_data = 2;
 
+  // Stacktrace that will be surfaced to the user.
+  // If this is provided this batch request is assumed to have failed.
   optional string exception_stack_trace = 3;
+
+  // Provided data from the backfila client telling backfila what part of the batch was not
+  // completed yet. Backfila will end up retrying retry either a new request with the remaining
```
swankjesse, line 110: retrying retry ---> I think you want to say “resuming” not retrying?

service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchAwaiter.kt
swankjesse: oh I think this is worse, just because there’s so much code between the declaration on line 46 and the value on line 100

swankjesse: might read better to extract this whole thing into a method?

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -589,6 +588,161 @@ class BackfillRunnerTest {
     }
   }
 
+  @Test fun `returning partial backfill bookmark makes followup runbatch`() {
```
swankjesse, line 591: nice test

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -589,6 +588,161 @@ class BackfillRunnerTest {
...
+        val nextBatch = fakeBackfilaClientServiceClient.runBatchRequests.receive()
+        // Batch ranges are no longer the same
+        assertThat(initialRequest.batch_range).isNotEqualTo(nextBatch.batch_range)
+      } finally {
+        runner.stop()
+      }
+    }
+  }
+
+  @Test fun `partial backfill bookmarks reduce parallelism`() {
```
swankjesse, line 654: maybe add a note about whether this is a desired or non-desired attribute of bookmarks

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -589,6 +588,161 @@ class BackfillRunnerTest {
...
+
+        waitForOtherCoroutines()
+        assertThat(fakeBackfilaClientServiceClient.runBatchRequests.poll()).isNull()
+      } finally {
+        runner.stop()
+      }
+    }
+  }
+
+  private suspend fun waitForOtherCoroutines() {
```
swankjesse, line 731: I don’t love this. Good enough for now, but it feels bad

swankjesse (approved):

client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -95,17 +95,23 @@ message RunBatchRequest {
   // User provided parameters at creation.
   map<string, bytes> parameters = 5;
   optional bool dry_run = 6;
-
```
mpawliszyn, line 98: Don't Merge!!!

client/src/main/proto/app/cash/backfila/client_service.proto
mpawliszyn: Backfila will end up resuming with a new request containing either the remaining range or the original range.

service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchAwaiter.kt
mpawliszyn: The good thing is that the robots are making sure I don't miss to return a value. Easy to understand... fair enough.

mpawliszyn: A method won't work because of the way it breaks out of the various loops. We can talk about it tomorrow during our pairing session perhaps?

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
mpawliszyn: Yeah agreed. It would have to be a separate refactor to get at the underlying coroutine scope.

client/src/main/proto/app/cash/backfila/client_service.proto
mpawliszyn: Ok fixed...

shellderp (approved):

service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchAwaiter.kt
```diff
@@ -57,6 +57,34 @@ class BatchAwaiter(
               backfillRunner.runBatchBackoff.addMillis(backfillRunner.metadata.extraSleepMs)
             }
           }
+
+          if (response.remaining_batch_range != null) {
+            // We have a remaining_batch_range, continue the batch.
+            val remaining = batch.newBuilder().batch_range(response.remaining_batch_range).build()
```
shellderp, line 63: do you have a test where the remaining batch fails and gets retried? I could see us regressing and retrying the original or something..

mpawliszyn: Good point

mpawliszyn: Ok added.

mpawliszyn: And it found a bug by the way... It did retry the original. No longer.

shellderp: oh damn I didn't even see it, nice

PR #158, opened 2021-01-13
Add Dockerfile to make integration testing with Backfila easier

jvmakine:
Also parametrise the db connection to make it easier to run the docker image against a containerised mysql.

Changes the jar main class from a non-existing `BackfilaServiceKt` to the development service `BackfilaDevelopmentServiceKt`


shellderp (approved):
neat

PR #166, opened 2021-03-12
make baseurl with subpath in it works to better support other clients

lyonlai:
what happened at the moment.

Take `prepareBackfill` as an example.

What happen now:
service url: `https://example.com/sub-path/` + `/backfila/prepare-backfil` (from Retrofit annotation) = `https://example.com/backfila/prepare-backfill`

After the change
service url: `https://example.com/sub-path/` + `backfila/prepare-backfil` (from Retrofit annotation) = `https://example.com/sub-path/backfila/prepare-backfill`

this will make it easy to integrate with other clients that mounts the backfila endpoint under different subpaths.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/client/HttpClientServiceApi.kt
+++ b/service/src/main/kotlin/app/cash/backfila/client/HttpClientServiceApi.kt
@@ -41,6 +41,6 @@ interface HttpClientServiceApi {
   ): RunBatchResponse
 
   companion object {
-    private const val BASE_PATH = "/backfila"
+    private const val BASE_PATH = "backfila"
   }
 }


tirsen (approved):

service/src/main/kotlin/app/cash/backfila/client/HttpClientServiceApi.kt
```diff
@@ -41,6 +41,6 @@ interface HttpClientServiceApi {
   ): RunBatchResponse
 
   companion object {
-    private const val BASE_PATH = "/backfila"
```
escardin, line 44: Shouldn't you have some other change potentially that ensures the base path has a trailing `/`? I would like to see some tests for this as well.

mpawliszyn: Yeah I'd like a couple tests. Particularly to prove that this won't break existing clients.

shellderp: may be hard to test since clients specify the path, but this is service code. It seems fine to me

mpawliszyn: Yeah true. As long as we check that existing clients are fine with the missing slash that will probably have to be enough.

Maybe try using the backfila Development app with existing misk Development app

chris-ryan-square (approved):

PR #170, opened 2021-03-18
SlackHelper messages indicate environment

zcross:


zcross:
Couldn't find any tests pertaining to this, understandably. Visual cues (like the web UI showing different colors per env) are useful, and the Slack messages are great but don't indicate environment at a glance (granted, the backfill links reveal this when actually visited)

shellderp (approved):
seems fine to me, I always prefer to use a separate channel per env as well

PR #172, opened 2021-03-26
Convert build from Groovy to KTS

swankjesse:


shellderp (approved):
never stamped so fast in my life

service/build.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+
+val jar by tasks.getting(Jar::class) {
+  manifest {
+    attributes("Main-Class" to "app.cash.backfila.service.BackfilaDevelopmentServiceKt")
+  }
+  isZip64 = true
+  classifier = "unshaded"
+}
+
+val shadowJar by tasks.getting(com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar::class) {
```
shellderp, line 80: do you know why we do shadow jar only here?

swankjesse: this is the deployable thing!

PR #174, opened 2021-04-21
Adding a bit more logging around shutting down a Backfill Runner.

mpawliszyn:


service/src/main/kotlin/app/cash/backfila/service/runner/BackfillRunner.kt
```diff
@@ -114,43 +114,48 @@ class BackfillRunner private constructor(
...
-        return@transaction true
+          return@transaction true
+        }
+        if (!dbRunning) break
+        delay(1000)
       }
-      if (!dbRunning) break
-      delay(1000)
+    } catch (e: Throwable) {
+      logger.info(e) { "Encountered an exception while monitoring the lease for ${logLabel()}" }
```
mpawliszyn, line 158: I think if we encounter an error here it is better to just catch and shut everything down, no?

shellderp: yeah, it will return out of this method and cancel the coroutines, right?

shellderp (approved):

service/src/main/kotlin/app/cash/backfila/service/runner/BackfillRunner.kt
mpawliszyn: Yeah it will return and run:
`coroutineContext.cancelChildren()`

PR #175, opened 2021-04-23
Adds a no-op task that runs before it tries to hunt for a lease.

mpawliszyn:
This will prevent a busy backfill node from attempting to take a lease.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/scheduler/RunnerSchedulerService.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/scheduler/RunnerSchedulerService.kt
@@ -37,6 +37,10 @@ class RunnerSchedulerService @Inject constructor(
 
   override fun run() {
     while (running) {
+      // once the noop runs there is space for another runner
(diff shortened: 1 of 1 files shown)


service/src/main/kotlin/app/cash/backfila/service/scheduler/RunnerSchedulerService.kt
```diff
@@ -37,6 +37,8 @@ class RunnerSchedulerService @Inject constructor(
 
   override fun run() {
     while (running) {
+      runnerExecutorService.submit { "noop task" }.get()
+      // once the noop runs there is space for another runner
       val newRunners = leaseHunter.hunt()
```
shellderp, line 42: should we check `if (!running) break` in case we just got shutdown, so we don't poll for another lease?

mpawliszyn: good call

shellderp (approved):

PR #182, opened 2021-05-03
Switch to runBlockingTest for test coroutines

shellderp:
This replaces delays with fake delays and now the coroutines run in the same scope so everything is deterministic. In some places we no longer need to wait since the test coroutine only wakes up when the runner coroutine has done its work

### DIFF
--- a/service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
+++ b/service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
@@ -22,15 +22,13 @@ import app.cash.backfila.protos.service.Parameter
 import app.cash.backfila.service.persistence.BackfilaDb
 import app.cash.backfila.service.persistence.BackfillState
 import app.cash.backfila.service.runner.BackfillRunner
+import app.cash.backfila.service.runner.EXTEND_LEASE_PERIOD
 import app.cash.backfila.service.scheduler.LeaseHunter
 import com.google.inject.Module
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.withTimeout
-import kotlinx.coroutines.withTimeoutOrNull
 import misk.hibernate.Transacter
 import misk.hibernate.load
 import misk.scope.ActionScope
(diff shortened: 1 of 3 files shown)


service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -355,7 +329,7 @@ class BackfillRunnerTest {
       }
 
       // Give enough time for the once-per-second check to cancel the jobs.
-      delay(1100)
+      delay(EXTEND_LEASE_PERIOD.toMillis())
```
mpawliszyn, line 332: plus a little bit?

mpawliszyn: Or perhaps half the lease period?

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -604,30 +552,23 @@ class BackfillRunnerTest {
...
+
+      delay(500)
+      // Nothing sent yet - the backoff is 1000ms
+      assertThat(fakeBackfilaClientServiceClient.runBatchRequests.poll()).isNull()
+      delay(500)
+      assertThat(fakeBackfilaClientServiceClient.runBatchRequests.poll()).isNotNull()
+      fakeBackfilaClientServiceClient.runBatchResponses.send(
+        Result.success(RunBatchResponse.Builder().build())
+      )
+      runner.stop()
```
mpawliszyn, line 571: stop doesn't have to be in the finally?

mpawliszyn (approved):
A few questions. But otherwise LGTM

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
shellderp: see the thing I made `runBlockingTestCancellable` it will cancel all children if any assert fails, so we no longer have to stop() in finally

service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
shellderp: doesn't have to be plus a bit, this consistently works

PR #189, opened 2021-05-31
Add names to coroutines

shellderp:
Helps debugging in tests

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
@@ -6,6 +6,7 @@ import app.cash.backfila.service.runner.BackfillRunner
 import com.google.common.base.Stopwatch
 import java.time.Duration
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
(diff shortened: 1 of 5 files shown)


service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
```diff
@@ -109,6 +110,7 @@ class BatchRunner(
       )
       rpcBackpressureChannel.upstream().send(Unit)
     }
+    rpcBackpressureChannel.upstream().close()
```
shellderp, line 113: technically this was missing but isn't really harmful either way..

mpawliszyn: Is this too aggressive? Will the awaiter be able to read from it after this still?

mpawliszyn (approved):
A small comment

service/src/main/kotlin/app/cash/backfila/service/runner/statemachine/BatchRunner.kt
shellderp: `close()` doesn't prevent reading all the messages already sent. the receive() will just error on ClosedException once all messages are read

`close(exception)` will immediately cancel the channel and the receive() fails. before this change we were just cancelling all the coroutines from the BackfillRunner anyway, now it just closes first then cancels. I don't feel strongly but it's just technically correct to close all the channels you send on

PR #193, opened 2021-06-15
Static Misk Datasource implementation

siwon-park:


client-misk-static/src/test/kotlin/app/cash/backfila/client/misk/static/StaticKotlinValBackfillTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    data class SauceAttributes(
+      val markSweet: Boolean = false,
+      val validate: Boolean = true
+    )
+
+    override val staticDatasource: List<String> = (sweetSauces + savourySauces).shuffled()
+  }
+
+  companion object {
+    val sweetSauces = listOf("Back-fil-a", "Polynesian", "Honey Mustard", "Barbeque", "Sriracha", "Honey Roasted BBQ")
```
mpawliszyn, line 88: Love the sauces

mpawliszyn (approved):

client-misk/src/main/kotlin/app/cash/backfila/client/misk/internal/EmbeddedBackfillRun.kt
```diff
@@ -124,7 +124,7 @@ internal class EmbeddedBackfillRun<B : Backfill>(
       )
     when (response.batches.isEmpty()) {
       true -> cursor.done = true
-      false -> cursor.previousEndKey = response.batches.map { it.batch_range.end }.maxOrNull()
+      false -> cursor.previousEndKey = response.batches.last().batch_range.end
```
mpawliszyn, line 127: It turns out a ByteString of '7' is > a ByteString of '10'. So we just do here what we do in the service and assume that the last batch is the the largest id.

PR #196, opened 2021-07-06
Refactoring to allow for a misk-less client

keeferrourke:


client-misk/src/main/kotlin/app/cash/backfila/client/embedded/EmbeddedBackfilaModule.kt
```diff
@@ -0,0 +1,17 @@
...
+import app.cash.backfila.client.BackfilaApi
+import app.cash.backfila.client.misk.BackfillModule
+import app.cash.backfila.client.internal.EmbeddedBackfila
+import com.google.inject.AbstractModule
+
+/**
+ * Use this to connect to an embedded Backfila implementation in development or testing. You will
+ * also need to install a [BackfillModule].
+ */
+class EmbeddedBackfilaModule : AbstractModule() {
```
shellderp, line 12: I was wondering about this. @swankjesse is there some standard KAbstractModule coming to Guice any time soon? Or should every project be defining its own copy of KAbstractModule, or just leave it this way?

client-misk/src/main/kotlin/app/cash/backfila/client/internal/BackfilaClient.kt
```diff
@@ -1,11 +1,11 @@
-package app.cash.backfila.client.misk.internal
+package app.cash.backfila.client.internal
```
shellderp, line 1: it looks like this file is still in the `client-misk` module. Should it be in `client`?

shellderp:
This is a great thing to do. if there's a long term vision here, would be great to discuss it in `#backfila-discuss` so we can be on the same page and I can help out

client-misk/src/main/kotlin/app/cash/backfila/client/embedded/EmbeddedBackfilaModule.kt
mpawliszyn: This is in progress but yeah we changed this so people didn't need the dependency here. Jesse might have better information

client-misk/src/main/kotlin/app/cash/backfila/client/internal/BackfilaClient.kt
keeferrourke: still very much WIP. @mpawliszyn and I worked on this for like 40 minutes this afternoon and didn't get too far 😛

service/src/main/kotlin/app/cash/backfila/service/runner/BackfillRunner.kt
```diff
@@ -108,27 +108,27 @@ class BackfillRunner private constructor(
   }
 
   @VisibleForTesting
-  fun start(coroutineScope: CoroutineScope) {
+  fun CoroutineScope.start() {
```
shellderp, line 111: I forget why I couldn't make this work. Seems good but wrong PR?

mpawliszyn: Yeah it turns out all of this is wrong according to @bezmax and @keeferrourke . I'll work with them in a followup to fix this more. Apparently we are blocking when we shouldn't on the main threads. There is also this really cool Flows thing that should work really really well.

shellderp: lets talk through that before big changes

client-base/README.md
```diff
@@ -0,0 +1,4 @@
+# client-base
+
+Contains the base implementation for a Backfila Client. Should only be depended on by Backfila\
```
shellderp, line 3: \

shellderp (approved):

PR #204, opened 2021-08-12
Quickfix: Switch EmbeddedBackfilaModule back to a KAbstractModule

keeferrourke:
EmbeddedBackfilaModule needs to be a `KAbstractModule` because misk idioms (e.g. `fun applicationModules(...): List<KAbstractModule>`) use a list of `KAbstractModule` for application modules, and `AbstractModule` does not fit those type constraints.


staktrace (approved):

keeferrourke:
We'll see about doing this another way.

PR #205, opened 2021-08-13
Expose client-base api through client-misk

keeferrourke:
Exposing `client-base` as an api of `client-misk` allows users of the client not to explicitly add both dependencies. `client-misk` will transitively pull in `client-base` and make its classes available to dependants.

i.e. to upgrade to the new backfila client rather than:

```kt
implementation(Dependencies.backfilaClientMisk)
implementation(Dependencies.backfilaClientBase)
```

you'll only require

```kt
implementation(Dependencies.backfilaClientMisk)
```

### DIFF
--- a/client-misk/build.gradle.kts
+++ b/client-misk/build.gradle.kts
@@ -19,7 +19,7 @@ dependencies {
   implementation(Dependencies.loggingApi)
 
   api(project(":client"))
-  implementation(project(":client-base"))
+  api(project(":client-base"))
 
   implementation(Dependencies.misk)
   implementation(Dependencies.miskActions)


shellderp:
What deps are you using? We designed client base to be internal implementation details that customers shouldn't need..

keeferrourke:
> What deps are you using? We designed client base to be internal implementation details that customers shouldn't need..

As of https://github.com/cashapp/backfila/pull/196, you need both artifacts on your classpath in order to resolve all the types.

This change makes it so that customers don't need to know about client-base... it just gets pulled in as an implementation detail of client-misk.

keeferrourke:
If client-base was meant to be wholly internal, that has inadvertently changed in #196.

keeferrourke:
Specifically, the BackfilaClientConfig moved to client-base. Maybe it should be in client?

keeferrourke:
#206 as an alternative to this PR

shellderp:
I see so at least those typealiases will not work as-is. Not sure what else might break. If we make this `api` though, all the internals get exposed to customers. `BackfilaClientConfig` does make sense to have in client since it should be visible to customers

keeferrourke:
Closing this in favour of #206!

PR #208, opened 2021-08-17
Expose client-misk through client-misk-dynamodb

keeferrourke:
Unlike `client-misk-{hibernate,jooq}`, the misk dynamodb client lists
`client-misk` as an implementation detail, and does not expose its API.
This appears to be error.

Following this change, dependents of `client-misk-dynamodb` will not need
to also explicitly depend on `client-misk`.


mpawliszyn (approved):

shellderp:
Long term I want dynamodb not to have any relation to misk (I mean, it already has basically none.. other than KAbstractModule) so this change doesn't really make sense to me

PR #214, opened 2021-09-24
Require injectable backfill classes when creating backfill modules

ktabouguia-sq:
Adding a simple safety check to backfill module to ensure that all
backfill classes are injectable at service startup. This helps us
avoid guice implementation exceptions when users try to run backfills
at runtime.


client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/DynamoDbBackfillModule.kt
```diff
@@ -20,6 +20,9 @@ class DynamoDbBackfillModule<T : DynamoDbBackfill<*, *>> private constructor(
 ) : KAbstractModule() {
   override fun configure() {
     install(DynamoDbBackfillBackendModule)
+    // Ensures that the backfill class has an @Inject annotation and
```
mpawliszyn, line 23: "Ensures that the backfill class is injectable. If you are failing this check you probably want to add an `@Inject` annotation to your class or check that all of your dependencies are provided."

I like it. I only propose the edit above since there are many ways your service might have to inject the Backfill.

mpawliszyn (approved):
As long as we address the comment, LGTM

client-misk-dynamodb/src/main/kotlin/app/cash/backfila/client/misk/dynamodb/DynamoDbBackfillModule.kt
ktabouguia-sq: Right, makes sense to me. Updated!

PR #219, opened 2021-10-06
Removing as much misk as possible from the backends.

mpawliszyn:


client-static/src/main/kotlin/app/cash/backfila/client/static/StaticDatasourceBackfillModule.kt
```diff
@@ -31,9 +35,13 @@ class StaticDatasourceBackfillModule<T : StaticDatasourceBackfill<*, *>> private
   }
 }
 
-private object StaticDatasourceBackfillBackendModule : KAbstractModule() {
+/**
+ * This is a kotlin object so these dependencies are only installed once.
+ */
+private object StaticDatasourceBackfillBackendModule : AbstractModule() {
```
keeferrourke, line 41: :+100000000:

client-misk-dynamodb/gradle.properties
```diff
@@ -1,4 +1,4 @@
 POM_ARTIFACT_ID=client-misk-dynamodb
 POM_NAME=client-misk-dynamodb
-POM_DESCRIPTION=Misk DynamoDB client library for backfila
+POM_DESCRIPTION=Deprecated Misk DynamoDB client library, use client-dynamodb instead.
```
keeferrourke, line 3: nice

client-misk/src/main/kotlin/app/cash/backfila/client/misk/Deprecations.kt
```diff
@@ -60,11 +60,11 @@ typealias NoParameters = NoParameters
...
 @Deprecated(
-    "Use MiskBackfillModule instead.",
-    replaceWith = ReplaceWith(
-        expression = "BackfillModule",
-        imports = ["app.cash.backfila.client.misk.MiskBackfillModule"]
-    ),
-    level = DeprecationLevel.ERROR
+  "Use MiskBackfillModule instead.",
+  replaceWith = ReplaceWith(
+    expression = "BackfillModule",
```
keeferrourke, line 65: I think expression is actually supposed to be `MiskBackfillModule`?

client-static/src/main/kotlin/app/cash/backfila/client/static/internal/StaticDatasourceBackfillOperator.kt
```diff
@@ -27,29 +25,18 @@ class StaticDatasourceBackfillOperator<I : Any, P : Any>(
...
-      throw BadRequestException("Start of range must be a number", e)
-    }
-
-    try {
-      end = request.range?.end?.utf8()?.toInt() ?: backfill.staticDatasource.size
-    } catch (e: NumberFormatException) {
-      throw BadRequestException("End of range must be a number", e)
-    }
+    val start = request.range?.start?.utf8()?.toInt() ?: 0
+    val end = request.range?.end?.utf8()?.toInt() ?: backfill.staticDatasource.size
```
keeferrourke, line 29: Do we want to use `toIntOrNull()` here instead of `toInt()`? That would take care of possible parse errors.

Gives us the option to just use defaults if we can't parse a number, or do something like the following without losing the messaging that would otherwise be propagated to the caller.

```kt
val startString = request.range?.start?.utf8() ?: "0"
val start = start.toIntOrNull() ?: error("Start of range must be a number")
```

client-static/src/test/kotlin/app/cash/backfila/client/static/StaticKotlinValBackfillTest.kt
```diff
@@ -97,11 +97,13 @@ class StaticKotlinValBackfillTest {
 
       this.assertThatCode {
         backfila.createWetRun<SaucesBackfill>(rangeStart = "abc")
-      }.hasMessageContaining("must be a number")
+      }.isInstanceOf(NumberFormatException::class.java)
+        .hasMessageContaining("For input string")
```
keeferrourke, line 101: Re my other comment, I think we can retain the information about start/end needing to be a number, rather than throwing some generic "couldn't parse input string" message

keeferrourke (approved):

PR #221, opened 2021-10-18
Finalize backfill

jclyne:
Adds a finalize hook to the backfill to allow for cleaning up resources , etc after the last batch

### DIFF
--- a/backfila-embedded/src/main/kotlin/app/cash/backfila/embedded/BackfillRun.kt
+++ b/backfila-embedded/src/main/kotlin/app/cash/backfila/embedded/BackfillRun.kt
@@ -32,9 +32,13 @@ interface BackfillRun<B : Backfill> {
 
   /** Prepares, scans and runs the whole backfill. */
   fun execute() {
-    precomputeRemaining()
-    scanRemaining()
-    runAllScanned()
+    try {
+      precomputeRemaining()
+      scanRemaining()
+      runAllScanned()
+    } finally {
+      finalizeBackfill()
+    }
     check(complete()) { "$this failed to run everything. Probably a bug." }
   }
 
@@ -60,6 +64,9 @@ interface BackfillRun<B : Backfill> {
 
   /** Runs all the batches that are scanned and ready to run. */
   fun runAllScanned()
+
+  /** Finalizes the Backfill */
+  fun finalizeBackfill()
 
   fun complete(): Boolean
 }
--- a/client-base/src/main/kotlin/app/cash/backfila/client/spi/BackfilaClientServiceHandler.kt
+++ b/client-base/src/main/kotlin/app/cash/backfila/client/spi/BackfilaClientServiceHandler.kt
@@ -3,6 +3,7 @@ package app.cash.backfila.client.spi
 import app.cash.backfila.client.BackfilaClientLoggingSetupProvider
 import app.cash.backfila.client.UnknownBackfillException
 import app.cash.backfila.client.internal.BackfillOperatorFactory
+import app.cash.backfila.protos.clientservice.FinalizeBackfillRequest
 import app.cash.backfila.protos.clientservice.GetNextBatchRangeRequest
 import app.cash.backfila.protos.clientservice.GetNextBatchRangeResponse
 import app.cash.backfila.protos.clientservice.PrepareBackfillRequest
@@ -79,6 +80,16 @@ class BackfilaClientServiceHandler @Inject constructor(
           .exception_stack_trace(ExceptionUtils.getStackTrace(exception))
           .build()
       }
+    }
+  }
+
+  @Throws(UnknownBackfillException::class)
+  fun finalizeBackfill(request: FinalizeBackfillRequest) {
+    loggingSetupProvider.withBackfillRunLogging(request.backfill_name, request.backfill_id) {
+      logger.info { "Finalizing backfill `${request.backfill_name}::${request.backfill_id}`" }
+
+      val operator = operatorFactory.create(request.backfill_name, request.backfill_id)
+      return@withBackfillRunLogging operator.finalizeBackfill()
     }
   }
 
--- a/client-dynamodb/src/main/kotlin/app/cash/backfila/client/dynamodb/internal/DynamoDbBackfillOperator.kt
+++ b/client-dynamodb/src/main/kotlin/app/cash/backfila/client/dynamodb/internal/DynamoDbBackfillOperator.kt
@@ -133,6 +133,14 @@ class DynamoDbBackfillOperator<I : Any, P : Any>(
       .build()
   }
 
+  override fun finalizeBackfill() {
+    try {
+      backfill.finalize()
+    } catch (t: Throwable) {
+      // Ignore errors from finalize
+    }
+  }
+
   private fun Map<String, AttributeValue>.toKeyRange(originalRange: DynamoDbKeyRange): KeyRange {
     require(originalRange.start + 1 == originalRange.end)
     return keyRangeCodec.encodeKeyRange(originalRange.start, originalRange.end, originalRange.count, this)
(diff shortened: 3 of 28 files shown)


service/src/main/kotlin/app/cash/backfila/service/runner/BackfillRunner.kt
```diff
@@ -99,7 +100,12 @@ class BackfillRunner private constructor(
   fun run() {
     factory.loggingSetupProvider.withLogging(backfillName, backfillRunId, partitionName) {
       runBlocking(MDCContext()) {
-        start(this)
+        try {
+          start(this)
+        } finally {
+          logger.info { "Runner finalizing backfill: ${logLabel()}" }
+          finalize()
```
shellderp, line 107: this isn't the end of the backfill. this can happen when the pod shuts down, or the backfill is paused. we need to do this in the code that sets it to COMPLETE state

client-static/src/main/kotlin/app/cash/backfila/client/static/StaticDatasourceBackfill.kt
```diff
@@ -49,6 +49,12 @@ abstract class StaticDatasourceBackfill<I : Any, P : Any> : Backfill {
   open fun runOne(item: I, config: BackfillConfig<P>) {
   }
 
+  /**
+   * Hook that gives you a place to finalize the backfill.
+   * This is called when the backfill is complete or errored, errors are ignored
```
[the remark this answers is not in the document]
shellderp: for every other RPC we retry if it failed (with backoff, and give up after n errors), I think we should be consistent. wouldn't it be bad if your finalize method doesn't get called? or if we have a network error

backfila-embedded/src/main/kotlin/app/cash/backfila/embedded/BackfillRun.kt
```diff
@@ -32,9 +32,13 @@ interface BackfillRun<B : Backfill> {
...
   fun execute() {
-    precomputeRemaining()
-    scanRemaining()
-    runAllScanned()
+    try {
+      precomputeRemaining()
+      scanRemaining()
+      runAllScanned()
+    } finally {
+      finalizeBackfill()
```
mpawliszyn, line 40: Yeah this is wrong. On error we shouldn't do this. Only on complete. MAYBE on marking the backfill cancelled.

mpawliszyn: We will need a variable to mark when this is complete. Also this is meant to be how people interact with the backfill. People should never call this. It is an internal implementation detail. complete will be true if the finalize returned a success.

client-base/src/main/kotlin/app/cash/backfila/client/spi/BackfilaClientServiceHandler.kt
```diff
@@ -82,6 +83,16 @@ class BackfilaClientServiceHandler @Inject constructor(
     }
   }
 
+  @Throws(UnknownBackfillException::class)
+  fun finalizeBackfill(request: FinalizeBackfillRequest) {
```
mpawliszyn, line 87: We need to pass in the request and make sure that it includes parameters, you may need them here too. Also we should create a response class.

client-dynamodb/src/main/kotlin/app/cash/backfila/client/dynamodb/internal/DynamoDbBackfillOperator.kt
```diff
@@ -133,6 +133,14 @@ class DynamoDbBackfillOperator<I : Any, P : Any>(
       .build()
   }
 
+  override fun finalizeBackfill() {
+    try {
+      backfill.finalize()
+    } catch (t: Throwable) {
+      // Ignore errors from finalize
```
mpawliszyn, line 140: Yeah no. You do this in your backfill if you need this to always catch.

client-dynamodb/src/test/kotlin/app/cash/backfila/client/dynamodb/DynamoDbTest.kt
```diff
@@ -54,6 +57,9 @@ class DynamoDbBackfillTest {
     dynamoDb: DynamoDBMapper
   ) : UpdateInPlaceDynamoDbBackfill<TrackItem, MakeTracksExplicitBackfill.ExplicitParameters>(dynamoDb) {
 
+    var finalized = false
```
mpawliszyn, line 60: Yeah this is logic that belongs in the embedded Run.

client-jooq/src/main/kotlin/app/cash/backfila/client/jooq/JooqBackfill.kt
```diff
@@ -85,6 +85,12 @@ abstract class JooqBackfill<K, Param : Any> : Backfill {
    */
   abstract fun backfill(backfillBatch: BackfillBatch<K, Param>)
 
+  /**
+   * Hook that gives you a place to finalize the backfill.
+   * This is called when the backfill is complete or errored, errors are ignored
```
mpawliszyn, line 90: Errors should not be ignored. If a customer wants to ignore errors they can do that themselves.

client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -131,3 +131,14 @@ message PipelinedData {
...
+
+message FinalizeBackfillRequest {
+  // A unique identifier for this backfill run.
+  // Can be used as a caching key since backfill metadata is immutable after creation.
+  optional string backfill_id = 1;
+
+  optional string backfill_name = 2;
+}
+
+message FinalizeBackfillResponse {
```
mpawliszyn, line 143: Yeah use this :)

client/src/main/proto/app/cash/backfila/client_service.proto
```diff
@@ -131,3 +131,14 @@ message PipelinedData {
 message PipelinedRecord {
   optional bytes data = 1;
 }
+
+message FinalizeBackfillRequest {
```
mpawliszyn, line 135: Make sure parameters are here too

```
  // User provided parameters at creation.
  map<string, bytes> parameters = 4;
```

service/src/main/kotlin/app/cash/backfila/client/BackfilaClientServiceClient.kt
```diff
@@ -13,4 +15,6 @@ interface BackfilaClientServiceClient {
   suspend fun getNextBatchRange(request: GetNextBatchRangeRequest): GetNextBatchRangeResponse
 
   suspend fun runBatch(request: RunBatchRequest): RunBatchResponse
+
+  suspend fun finalizeBacfkill(request: FinalizeBackfillRequest): FinalizeBackfillResponse
```
mpawliszyn, line 19: typo :)

service/src/main/kotlin/app/cash/backfila/service/runner/BackfillRunner.kt
mpawliszyn: We need a couple tests around this too.

service/src/test/kotlin/app/cash/backfila/client/FakeBackfilaClientServiceClient.kt
```diff
@@ -91,4 +93,7 @@ class FakeBackfilaClientServiceClient @Inject constructor() : BackfilaClientServ
     runBatchRequests.send(request)
     return runBatchResponses.receive().getOrThrow()
   }
+
+  override suspend fun finalizeBacfkill(request: FinalizeBackfillRequest) =
```
mpawliszyn, line 97: We should do something similar to the Prepare and channel the requests. Then we can use this to test the service.

PR #226, opened 2021-11-06
Upgrade misk and kotlin

shellderp:
Upped kotlin coroutines to match kotlin version, which changed some experimental APIs in VariableCapacityChannel


.circleci/config.yml
```diff
@@ -11,12 +11,12 @@ jobs:
     steps:
       - checkout
       - run:
-          name: "Download and extract OpenJDK 11"
+          name: "Download and extract OpenJDK 14"
           command: |
-            mkdir -p ~/openjdk-11 && cd ~/openjdk-11 && curl -sSL https://download.oracle.com/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz | tar -xz --strip-components 1
+            mkdir -p ~/openjdk-14 && cd ~/openjdk-14 && curl -sSL https://download.java.net/java/GA/jdk14.0.2/205943a0976c4ed48cb16f1043c5c647/12/GPL/openjdk-14.0.2_linux-x64_bin.tar.gz | tar -xz --strip-components 1
```
keeferrourke, line 16: Why update to JDK 14? It's not an LTS release.

Would it be preferred to wait until there's good kotlin/gradle support for JDK 17 and then update to that?

keeferrourke: Also this feels like something that could be hermitized  https://cashapp.github.io/hermit/usage/get-started/

shellderp: misk was built with jdk 14, for whatever reason. should I build misk with jdk 11 instead?

shellderp: hah you guys rebuilt nix basically? I tried also using circleci docker images instead, which nearly worked but got hung up on the vitess testing framework that manually talks to docker

keeferrourke: > misk was built with jdk 14, for whatever reason. so this was necessary to get the artifact. should I build misk with jdk 11 instead?

if this is for misk compat, I'd just do what misk does. can see what upstream says about publishing with java 11 vs 14 and adapt later

shellderp: reverting this to 11

keeferrourke (approved):

PR #229, opened 2021-11-21
Support strings as keys, and rework how we handle primary key types

shellderp:
- Don't assume the pkey.toString() is reliable, define an adapter for both from and to ByteString
- Allow binding more types
- Let Hibernate + prepared statements do the work for us when it comes to querying the primary key


client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/IdPrimaryKeyCursorAdapter.kt
```diff
@@ -0,0 +1,18 @@
...
+import misk.hibernate.Id
+import okio.ByteString
+import okio.ByteString.Companion.encodeUtf8
+
+internal object IdPrimaryKeyCursorAdapter : PrimaryKeyCursorAdapter<Id<*>> {
+  override fun toByteString(pk: Id<*>): ByteString {
+    return pk.toString().encodeUtf8()
+  }
+
+  override fun fromByteString(byteString: ByteString): Result<Id<*>> {
```
shellderp, line 13: why Result: It forces you to check, and gives an obvious way to fail

client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/internal/HibernateBackfillOperator.kt
```diff
@@ -196,8 +195,12 @@ internal class HibernateBackfillOperator<E : DbEntity<E>, Pkey : Any, Param : An
         val stopwatch = Stopwatch.createStarted()
         boundingMax = boundingRangeStrategy
           .computeBoundingRangeMax(
-            backfill, partitionName, previousEndKey, backfillRange,
-            scanSize
+            backfill,
+            partitionName,
+            previousEndKey?.let { primaryKeyCursorMapper.fromByteString(backfill.pkeyClass.java, it).getOrThrow() },
```
shellderp, line 200: we can probably store these as PKey instead of strings, will come back to it

client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/IdPrimaryKeyCursorAdapter.kt
mpawliszyn: I like it.

client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/internal/BoundingRangeStrategy.kt
```diff
@@ -104,29 +130,52 @@ private fun <E : DbEntity<E>, Pkey : Any> selectMaxBound(
...
   var where = when {
-    previousEndKey != null -> "WHERE $pkeyName > ${previousEndKey.utf8()}"
-    else -> "WHERE $pkeyName >= ${backfillRange.start.utf8()}"
+    previousEndKey != null -> {
+      params.add(previousEndKey)
+      "WHERE $pkeyName > ?"
+    }
+    else -> {
+      params.add(backfillRangeStart)
+      "WHERE $pkeyName >= ?"
```
mpawliszyn, line 149: wouldn't named parameters be be better? The '?' makes me nervous.

client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/internal/HibernateBackfillOperator.kt
```diff
@@ -196,8 +191,12 @@ internal class HibernateBackfillOperator<E : DbEntity<E>, Pkey : Any, Param : An
         val stopwatch = Stopwatch.createStarted()
         boundingMax = boundingRangeStrategy
           .computeBoundingRangeMax(
-            backfill, partitionName, previousEndKey, backfillRange,
-            scanSize
+            backfill,
+            partitionName,
+            previousEndKey,
+            backfillRange.start.let { primaryKeyCursorMapper.fromByteString(backfill.pkeyClass.java, it).getOrThrow() },
```
mpawliszyn, line 197: maybe the class should expose a conversion method instead. I'd think the mapper and the backfill should continue to be related.

mpawliszyn (approved):
Some minor stuff but looks good overall.

client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/internal/BoundingRangeStrategy.kt
shellderp: I can't find a way to use named params and get the right return type. Have to use a `PreparedStatement` to use the nullSafeGet later on, and it doesn't support naming like createNativeQuery does

client-misk-hibernate/src/main/kotlin/app/cash/backfila/client/misk/hibernate/internal/HibernateBackfillOperator.kt
shellderp: The type is independent of the backfill, just like how it's bound in hibernate independent of the backfill

PR #232, opened 2022-01-15
GetBackfillRunsAction: Summarize partition data with one query

shellderp:
This way we don't try to load thousands into memory

Also added tests as this was untested.


service/src/main/kotlin/app/cash/backfila/dashboard/GetBackfillRunsAction.kt
```diff
@@ -120,27 +118,42 @@ class GetBackfillRunsAction @Inject constructor(
...
+  private fun partitionSummary(
     session: Session,
-    pausedBackfills: List<DbBackfillRun>
-  ) = queryFactory.newQuery<RunPartitionQuery>()
-    .backfillRunIdIn(pausedBackfills.map { it.id })
-    .list(session)
-    .groupBy { it.backfill_run_id }
+    runs: List<DbBackfillRun>,
+  ): Map<Id<DbBackfillRun>, PartitionSummary> {
+    val list = session.hibernateSession.createQuery("""
```
shellderp, line 131: wish sqldelight was set up or misk queries could support it, but whatever this works

PR #234, opened 2022-02-07
Add API for finalize and client stub

shellderp:
Just adding a stub so we can deploy this on clients and the service can start calling it without 404ing

Service changes will come later - not adding the operator functions yet so people don't implement this in their backfills expecting it to work


shellderp:
something is broken with circleci

client-base/src/main/kotlin/app/cash/backfila/client/spi/BackfilaClientServiceHandler.kt
```diff
@@ -82,6 +83,16 @@ class BackfilaClientServiceHandler @Inject constructor(
     }
   }
 
+  @Throws(UnknownBackfillException::class)
+  fun finalizeBackfill(request: FinalizeBackfillRequest): FinalizeBackfillResponse {
+    return loggingSetupProvider.withBackfillRunLogging(request.backfill_name, request.backfill_id) {
+      logger.info { "Finalizing backfill `${request.backfill_name}::${request.backfill_id}`" }
```
mpawliszyn, line 89: We should probably push the stub into the operator factory?

mpawliszyn (approved):
I think the stub should be at the OperatorFactory level no?

PR #241, opened 2022-03-16
Adding Backfila functions for easy Java usage.

mpawliszyn:


backfila-embedded/src/main/kotlin/app/cash/backfila/embedded/Backfila.kt
```diff
@@ -33,25 +33,110 @@ inline fun <reified Type : Backfill> Backfila.createDryRun(
...
+//**************************************************
+// Dry Run Signatures
+fun <Type : Backfill> Backfila.createDryRun(
+  backfill: Class<Type>,
+): BackfillRun<Type> {
+  return createDryRun(backfill.kotlin, null, mapOf(), null, null)
+}
+
+@JvmOverloads
+fun <Type : Backfill> Backfila.createDryRun(
```
keeferrourke, line 98: So this works, but I think it won't be as nice for Java callers as we thought. See https://stackoverflow.com/questions/28294509/accessing-kotlin-extension-functions-from-java

We'd end up having to call something like:

```java
@Inject Backfila backfila
...
BackfilaKt.createDryRun(backfila, MyBackfill.class)
```

which isn't great.

shellderp (approved):

mpawliszyn:
Ended up creating something called JavaBackfila to be Java's equivalent to the Kotlin Backfila object. Both these objects represent the customer's use of the UI in tests.

keeferrourke (approved):

PR #246, opened 2022-04-06
Fix maven publish plugin

shellderp:
wire duplicates proto files to another directory and gradle was unhappy there were duplicate paths. Not sure why this only started recently


mpawliszyn (approved):
@keeferrourke might have an opinion.

shellderp:
I will merge since it fixes master. Can make further improvements after

keeferrourke:
LGTM, thanks!

PR #254, opened 2022-05-11
Allow BackfilaClientConfig to be constructed late.

mpawliszyn:
We may not always want the config to be provided up front for all integrations so this allows for late construction.


shellderp:
could you allow the user to bind a provider instead? that's what I did in our internal version

PR #256, opened 2022-05-28
Do not sleep after empty/fake batch runs

shellderp:
When using extra_sleep with a big sparse table, it runs super slow for no good reason

This is because we still "run" empty batches. We don't actually send an RPC, but we forward a fake response to the Awaiter where the cursors can get pushed forward. Otherwise we'd lose a lot of progress on a restart

### DIFF
--- a/service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
+++ b/service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
@@ -590,7 +590,7 @@ class BackfillRunnerTest {
     }
   }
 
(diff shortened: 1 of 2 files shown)


service/src/test/kotlin/app/cash/backfila/service/BackfillRunnerTest.kt
```diff
@@ -620,6 +620,74 @@ class BackfillRunnerTest {
     }
   }
 
+  @Test fun `extra sleep with 2 threads only sleeps once`() {
```
shellderp, line 623: unrelated bonus test of existing code

mpawliszyn (approved):
LGTM

PR #258, opened 2022-06-05
Add database entity for parameters

shellderp:
Start dual writing to a new table. Next PR will be a self backfill which should be fun
Will also need another table for run_parameters

### DIFF
--- a/client/src/main/proto/app/cash/backfila/service.proto
+++ b/client/src/main/proto/app/cash/backfila/service.proto
@@ -10,6 +10,7 @@ message Parameter {
   optional string name = 1;
   // Description to show in the UI.
   optional string description = 2;
+  optional bool required = 3;
 }
 
 message ConfigureServiceRequest {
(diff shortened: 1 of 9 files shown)


client/src/main/proto/app/cash/backfila/service.proto
```diff
@@ -10,6 +10,7 @@ message Parameter {
   optional string name = 1;
   // Description to show in the UI.
   optional string description = 2;
+  optional bool required = 3;
```
shellderp, line 13: Not sure how to wire this in the client yet. It's just so we can indicate in the web UI when a param is required

mpawliszyn: It can just be part of the Annotation?

shellderp: well in kotlin we know if it has a default or not, I just don't know where that is determined today. (won't be hard)

service/src/main/kotlin/app/cash/backfila/service/persistence/DbRegisteredBackfill.kt
```diff
@@ -54,9 +57,14 @@ class DbRegisteredBackfill() : DbUnsharded<DbRegisteredBackfill>, DbTimestampedE
   override lateinit var updated_at: Instant
 
   // TODO(mgersh): this might want to be its own table
```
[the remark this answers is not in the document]
shellderp: when I remove it yep

PR #263, opened 2022-07-06
For better startup errors in tests.

mpawliszyn:
I think we only want to skip failing on startup when it is a real
external client. Otherwise we want to fail to make the true error move
visible.


mpawliszyn:
I encountered a failing startup and the failure in tests looks like it was never called. This will help people who have failing configuration that the EmbeddedBackfila is trying to warn them about.

shellderp (approved):

PR #271, opened 2022-08-09
[WIP] Implement support for Google Spanner

jdm-square:
These changes add a new backend for backfilling Spanner databases integrated into misk services.

I'm still adding unit tests to show that it all works, but I figured I would put it up for some early review.


client-base/src/main/kotlin/app/cash/backfila/client/RealBackfillModule.kt
```diff
@@ -21,8 +21,8 @@ class RealBackfillModule(
   constructor(
     config: BackfilaClientConfig,
     loggingSetupProvider: KClass<out BackfilaClientLoggingSetupProvider> =
-        BackfilaClientNoLoggingSetupProvider::class
-  ) : this( Provider { config }, loggingSetupProvider)
+      BackfilaClientNoLoggingSetupProvider::class
```
deanpapastrat, line 24: Nit: remove whitespace changes from PR so you don't end up on a `git blame` by accident

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/SpannerBackfillModule.kt
```diff
@@ -0,0 +1,70 @@
...
+ * This is a kotlin object so these dependencies are only installed once.
+ */
+private object SpannerBackfillBackendModule : KAbstractModule() {
+  override fun configure() {
+    Multibinder.newSetBinder(binder(), BackfillBackend::class.java).addBinding()
+      .to(SpannerBackend::class.java)
+  }
+
+  @Provides @Singleton @ForSpannerBackend
+  fun provideDynamoMoshi(): Moshi {
```
deanpapastrat, line 55: Nit: naming

```suggestion
  fun provideSpannerMoshi(): Moshi {
```

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/SpannerBackfill.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import app.cash.backfila.client.Backfill
+import app.cash.backfila.client.BackfillConfig
+import com.google.cloud.spanner.DatabaseClient
+import com.google.cloud.spanner.KeyRange
+
+abstract class SpannerBackfill<Param : Any> : Backfill {
+  abstract val dbClient: DatabaseClient
+  abstract val tableName: String
+  abstract val primaryKeyColumns: List<String>
```
deanpapastrat, line 11: Nit: I really like when we add documentation to these fields. It makes it much better in the future.

Also, I did realize recently that I think we could derive the primary key columns from the table name. IDK if that matters to you or not.

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackend.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlin.reflect.full.findAnnotation
+
+@Singleton
+internal class SpannerBackend @Inject constructor(
+  private val injector: Injector,
+  @ForSpannerBackend private val backfills: MutableMap<String, KClass<out SpannerBackfill<*>>>,
+  @ForSpannerBackend internal val moshi: Moshi,
+  internal val spanner: Spanner,
+  // internal var primaryKeyCursorMapper: PrimaryKeyCursorMapper,
+  // internal var queryFactory: Query.Factory,
```
deanpapastrat, line 30: Nit: cut out commented code

```suggestion
```

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackend.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import java.lang.reflect.ParameterizedType
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+import kotlin.reflect.full.findAnnotation
+
+@Singleton
+internal class SpannerBackend @Inject constructor(
```
deanpapastrat, line 24: General observation: this looks _almost_ entirely like something that could be made generic across backfills.

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
```diff
@@ -0,0 +1,108 @@
...
+      KeySet.range(
+        KeyRange.openClosed(
+          Key.of(*previousEndKey.toTypedArray()),
+          Key.of(),
+        )
+      )
+    }
+
+    val query = backfill.dbClient.singleUseReadOnlyTransaction()
+      .read(backfill.tableName, range, backfill.primaryKeyColumns)
```
deanpapastrat, line 61: I believe you can pass options to this query. You should make it a "low priority" query - otherwise a backfill which runs at "max speed" could overwhelm the DB. Plus we want to tell Spanner that backfill traffic isn't as important.

client-misk-spanner/src/test/kotlin/app/cash/backfila/client/misk/spanner/SpannerBackfillTest.kt
```diff
@@ -0,0 +1,161 @@
...
+        .setColumnTo(TrackData.Column.album_token.name, "token")
+        .setColumnTo(TrackData.Column.track_title.name, "track title")
+        .setColumnTo(TrackData.Column.artist_name.name, "artist")
+        .build()
+      it.buffer(mutation)
+    }
+  }
+
+  @AfterEach
+  fun cleanup() {
```
deanpapastrat, line 63: I wonder if I should port some of the grantly helpers into miskGcp...

deanpapastrat (approved):
LGTM!

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
```diff
@@ -0,0 +1,108 @@
...
+  override fun name() = backfill.javaClass.toString()
+
+  override fun prepareBackfill(request: PrepareBackfillRequest): PrepareBackfillResponse {
+    val config =
+      parametersOperator.constructBackfillConfig(request.parameters, request.dry_run)
+    backfill.validate(config)
+
+    val partitions = listOf(
+      PrepareBackfillResponse.Partition.Builder()
+        .backfill_range(request.range)
```
shellderp, line 37: are you requiring range to be passed in? In other implementations we compute the ranges if you don't pass it in

client-misk-spanner/src/test/kotlin/app/cash/backfila/client/misk/spanner/SpannerBackfillTest.kt
jdm-square: It would certainly reduce some of the need to reinvent the wheel in other projects.

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/SpannerBackfill.kt
jdm-square: Can you elaborate on how deriving the primary key columns would work? Would it require reading from the spanner schema information tables or something?

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
jdm-square: As far as I can tell, options can only be passed for transactions that can perform writes.

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
jdm-square: The range is actually completely ignored. Spanner is unlike many other DBs, where for optimal performance primary keys really can't be in anything like a monotonic increasing range. I don't know how to compute a range without doing a full table scan, which seems... suboptimal.

shellderp: you can't ask for min/max primary key value?

jdm-square: Primary keys are often random values like UUIDs and unordered for optimal performance. Min/max aren't valid concepts, as far as I can tell.

shellderp: Backfila requires ordered key values to operate. I'm curious how you would use it if that's not the case. I haven't used spanner but my understanding was its ordered, you just want to avoid sequential writes

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/SpannerBackfill.kt
jdm-square: I think you're talking about https://cloud.google.com/spanner/docs/information-schema#index_columns, and that feels like more effort than I really want right now.

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
deanpapastrat: So the keys are ordered - but other keys could be added inside your key range during the backfill since they aren’t guaranteed to be an auto-incrementing integer.

From the spanner docs: “ Cloud Spanner stores rows in sorted order by primary key values”.

The reason you want well-distributed keys is because it helps prevent “hot shards” in a distributed system, which it has in common with dynamo. Therefore, I believe this should be fine.

deanpapastrat: And to answer the original question - we don’t require a range to be passed in. That’s optional.

deanpapastrat (approved):
LGTM

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
```diff
@@ -0,0 +1,117 @@
...
+        .build()
+    )
+
+    return PrepareBackfillResponse.Builder()
+      .partitions(partitions)
+      .build()
+  }
+
+  override fun getNextBatchRange(request: GetNextBatchRangeRequest): GetNextBatchRangeResponse {
+    // Establish a range to scane - either we want to start at the first key,
```
shellderp, line 48:
```suggestion
    // Establish a range to scan - either we want to start at the first key,
```

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
```diff
@@ -0,0 +1,117 @@
...
+    // or start from (and exclude) the last key that was scanned.
+    val range = if (request.previous_end_key == null) {
```
shellderp, line 50: I guess we're not using the backfill_range at all, that's what would be passed in by the user (or I missed it somewhere)

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
shellderp: Yeah I'm well aware how primary key design works in spanner, and you can have items added within the range. That's true even in auto increment, technically. It doesn't matter since the expectation is you are inserting new items that don't need backfilling.

It sounds like you are able to just ask spanner for records and it will give in some order, that should be fine I guess

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
jdm-square: Yes. If I'm not mistaken, the DynamoDB backend also ignores it.

shellderp: DynamoDb is pretty limited because of dynamo itself, the hibernate one is pretty good to copy from. Obviously, build whatever features you want, I won't be using it :P

client-misk-spanner/src/main/kotlin/app/cash/backfila/client/misk/spanner/internal/SpannerBackfillOperator.kt
escardin: Wouldn't this work like dynamo backfills? Dynamo is somewhat different, but has a scan mechanism we use, and I believe we don't do ranges on it either? You could check that.

escardin: Does this mean that you will essentially run your backfill single threaded?

escardin (commented):
Since this isn't urgent, I'll just comment.

I think this is a good start, but we'd want to really make sure as much of backfila works as expected as possible. Having a single partition is okay, but could be somewhat challenging to scale. Upper and lower bounds for ranges are a reasonable tradeoff.

mpawliszyn (commented):
Overall looking very good. Let's avoid misk except in test.

I wonder if you can use this to be more parallel?
https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Spanner.PartitionRead
Can you share a session among different machines? Your backfill might die if the session dies though.

PR #273, opened 2022-09-15
Changing how throwOnStartup works to fix the retrofit client.

mpawliszyn:


keeferrourke (approved):
This is a lot nicer 👍

shellderp (approved):

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -82,8 +82,8 @@ object Versions {
   val flywayDBPlugin = "5.2.4"
   val jooqGenPlugin = "5.2"
   val jooq = "3.14.8"
-  val kotlin = "1.5.31"
-  val kotlinCoroutines = "1.5.2"
+  val kotlin = "1.7.0"
```
shellderp, line 85: is it needed in this PR?

keeferrourke: I think this is breaking the build actually?

```
* What went wrong:
A problem occurred configuring project ':buildSrc'.
> Could not resolve all files for configuration ':buildSrc:classpath'.
   > Could not resolve org.jetbrains.kotlin:kotlin-gradle-plugin-idea:1.7.0.
```

mpawliszyn: That was the original error that was breaking the build. This was an attempt to fix that.
https://app.circleci.com/pipelines/github/cashapp/backfila/1804/workflows/5819c266-db5e-4f68-9499-ad5cb0912902/jobs/4623

PR #276, opened 2022-09-20
Remove jcenter as a repository

staktrace:
jcenter has been shut down and the build fails to download things from there.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -66,7 +66,6 @@ subprojects {
 
   repositories {
     mavenCentral()
-    jcenter()
     maven(url = "https://s3-us-west-2.amazonaws.com/dynamodb-local/release")
   }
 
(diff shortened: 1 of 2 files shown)


mpawliszyn (approved):

staktrace:
From https://blog.gradle.org/jcenter-shutdown:
```
The Gradle Plugin Portal implicitly mirrors JCenter currently. If you’re using the Plugin Portal (via gradlePluginPortal() or the URL plugins.gradle.org/m2) to resolve your application’s dependencies, you may be relying on JCenter. You should avoid using the Plugin Portal as a repository, except for Gradle plugin projects.
```

mpawliszyn:
> From https://blog.gradle.org/jcenter-shutdown:
> 
> ```
> The Gradle Plugin Portal implicitly mirrors JCenter currently. If you’re using the Plugin Portal (via gradlePluginPortal() or the URL plugins.gradle.org/m2) to resolve your application’s dependencies, you may be relying on JCenter. You should avoid using the Plugin Portal as a repository, except for Gradle plugin projects.
> ```

I think we still need the plugin portal but I guess we place it second?

PR #277, opened 2022-10-28
Set a limit to the number of run partitions retrieved

tgregory-block:
Places a limit on the number of unleased partitions to retrieve, to ensure we don't trip the maximum row limit in misk's hibernate module. Went with a lower batch size since we ultimately select one at random

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/service/scheduler/LeaseHunter.kt
+++ b/service/src/main/kotlin/app/cash/backfila/service/scheduler/LeaseHunter.kt
@@ -29,7 +29,7 @@ class LeaseHunter @Inject constructor(
         .runState(BackfillState.RUNNING)
         .leaseExpiresAtBefore(clock.instant())
         .apply {
-          maxRows = RUN_PARTITION_BATCH_SIZE
+          maxRows = RUN_PARTITION_QUERY_LIMIT
         }
         .list(session)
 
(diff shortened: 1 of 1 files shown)


shellderp (approved):

service/src/main/kotlin/app/cash/backfila/service/scheduler/LeaseHunter.kt
```diff
@@ -54,5 +57,6 @@ class LeaseHunter @Inject constructor(
 
   companion object {
     val LEASE_DURATION: Duration = Duration.ofMinutes(5)
+    const val RUN_PARTITION_BATCH_SIZE = 100
```
mpawliszyn, line 60: RUN_PARTITION_QUERY_LIMIT ? This should be more than 100. At least 1000. Can we set it to 10,000 and call it a day?

shellderp: this appears to limit how many unleased partitions are fetched to try to lease them, it doesn't limit partitions per run

tgregory-block: If we're selecting a single random partition, why does the size of the pool matter?

tgregory-block: It seems we could even set `maxRows = 1`?

shellderp: maxrows=1 would mean all pods select the same one, the order is deterministic. I think 100 was fine..

mpawliszyn: It will randomly choose over a larger set. I guess 100 is ok too. I appreciate the name change as we are doing no batching here.

mpawliszyn (approved):

PR #283, opened 2022-12-05
Stop using backfill_id in Prepare request and deprecate it

shellderp:
We actually send the registered_backfill_id here, which is not what the docs say at all and misleading. We don't know the run id until Prepare gets a response. We should remove it

### DIFF
--- a/client-misk/src/main/kotlin/app/cash/backfila/client/misk/client/BackfilaClientMDCLoggingSetupProvider.kt
+++ b/client-misk/src/main/kotlin/app/cash/backfila/client/misk/client/BackfilaClientMDCLoggingSetupProvider.kt
@@ -8,10 +8,12 @@ import wisp.logging.getLogger
 class BackfilaClientMDCLoggingSetupProvider @Inject constructor() :
   BackfilaClientLoggingSetupProvider {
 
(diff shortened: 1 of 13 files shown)


client-misk/src/main/kotlin/app/cash/backfila/client/misk/client/BackfilaClientMDCLoggingSetupProvider.kt
```diff
@@ -8,10 +8,12 @@ import wisp.logging.getLogger
 class BackfilaClientMDCLoggingSetupProvider @Inject constructor() :
   BackfilaClientLoggingSetupProvider {
 
-  override fun <T> withBackfillRunLogging(backfillName: String, backfillId: String, wrapped: () -> T): T {
+  override fun <T> withBackfillRunLogging(backfillName: String, backfillId: String?, wrapped: () -> T): T {
     try {
       MDC.put(MDC_BACKFILL_NAME, backfillName)
-      MDC.put(MDC_BACKFILL_ID, backfillId)
+      backfillId?.let {
```
mpawliszyn, line 14: We should probably just combine the two methods?

mpawliszyn (approved):

PR #289, opened 2023-01-30
Move test CI from CircleCI to Github Actions

adrw:
Given recent CircleCI security issues, this aligns Backfila repo with most other CashApp repos which are already on Github Actions. A follow up PR will remove the CircleCI config but for now we will dual-run CI.


bin/jaotc
```diff
@@ -0,0 +1 @@
+.openjdk@11.pkg
```
shellderp, line 1: are these bin files intentional?

adrw: Yes, they're part of the Hermit config and meant to be checked in. For example in Misk: https://github.com/cashapp/misk/tree/master/bin

shellderp: ok it's not obvious at a glance I guess, this works better than a file listing deps?

bin/activate-hermit
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# This file must be used with "source bin/activate-hermit" from bash or zsh.
```
shellderp, line 2: so I run this to set up my environment after pulling this change?

shellderp: the home README.md should be updated to say how you get dev env set up with hermit

adrw: https://github.com/cashapp/backfila/pull/291

bin/jaotc
adrw: Yes, I initially was hesitant but over the past year it's made toolchain management a lot simpler across internal and external repos.

.github/workflows/Test.yml
```diff
@@ -0,0 +1,76 @@
+name: Test
```
adrw, line 1: File copied from Misk with Backfila DB docker line added

build.gradle.kts
```diff
@@ -115,3 +115,8 @@ subprojects {
     }
   }
 }
+
+// Disable the Gradle wrapper if Gradle is managed by Hermit
```
adrw, line 119: Missed adding this in the Hermit PR #291

danieloh0714 (approved):

r3mariano (approved):

shellderp (approved):
oke

PR #291, opened 2023-02-01
Use Hermit for JDK and Gradle setup

adrw:


bin/README.hermit.md
```diff
@@ -0,0 +1,7 @@
+# Hermit environment
```
adrw, line 1: Auto-generated and supposed to be checked-in

PR #297, opened 2023-03-14
Pass backfill_id in EmbeddedBackfila

shellderp:
More realistic


mpawliszyn (approved):
Should it be a string or something else?

PR #305, opened 2023-04-13
Ignore empty lines in S3 by default

shellderp:


client-s3/src/main/kotlin/app/cash/backfila/client/s3/record/Utf8StringNewlineStrategy.kt
```diff
@@ -6,8 +6,16 @@ import okio.ByteString
 /**
  * Produces single string records separated by newlines.
  */
-class Utf8StringNewlineStrategy : RecordStrategy<String> {
+class Utf8StringNewlineStrategy(
+  private val ignoreBlankLines: Boolean = true,
+) : RecordStrategy<String> {
   override fun calculateNextRecordBytes(source: BufferedSource): Long = StringStrategyTools.newlineByteCalculator(source)
```
mpawliszyn, line 12: Don't we want to use a slightly different next record strategy too?

mpawliszyn:
I created a string strategy for this too here:
https://github.com/cashapp/backfila/pull/320/files

So this PR is no longer necessary.

PR #307, opened 2023-04-24
Prepare backfill: catch exception in clients and send message

shellderp:
Will only merge this after we have all deployed the new server code


mpawliszyn (approved):
This is great!

PR #313, opened 2023-05-12
Adding scan byte chunking strategies for GetNextBatchRange.

mpawliszyn:
I had the semantics wrong. Not putting in an end assumes you want the whole file. So I'm adding an adaptive strategy to find the chunk size.
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/GetObjectRequest.html#setRange-long-

I will add some tests before merging but sending this out as it is for comment.

### DIFF
--- a/client-s3/src/main/kotlin/app/cash/backfila/client/s3/internal/S3DatasourceBackfillOperator.kt
+++ b/client-s3/src/main/kotlin/app/cash/backfila/client/s3/internal/S3DatasourceBackfillOperator.kt
@@ -109,22 +109,22 @@ class S3DatasourceBackfillOperator<R : Any, P : Any>(
       ).build()
     }
 
-    val fileStream = s3Service.getFileStreamStartingAt(
-      backfill.getBucket(config),
-      pathPrefix + request.partition_name,
-      previousEndKey,
+    val bytesToScan = Math.min(previousEndKey + backfill.scanByteStrategy.bytesToScan(), fileSize)
+    val fileStream = s3Service.getFileChunkSource(
+        backfill.getBucket(config),
+        pathPrefix + request.partition_name,
+        previousEndKey,
+        bytesToScan,
     )
 
     val result = fileStream.use {
       val recordBytes = mutableListOf<Long>()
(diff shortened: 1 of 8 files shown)


client-s3/src/main/kotlin/app/cash/backfila/client/s3/internal/S3DatasourceBackfillOperator.kt
```diff
@@ -109,22 +109,22 @@ class S3DatasourceBackfillOperator<R : Any, P : Any>(
...
           (
-              request.compute_time_limit_ms == null || // Either there is no limit or we are withing our timeframe.
-                  stopwatch.elapsed(TimeUnit.MILLISECONDS) <= request.compute_time_limit_ms
-              )
+              request.compute_time_limit_ms == null || // Either there is no limit or we are within our timeframe.
+              stopwatch.elapsed(TimeUnit.MILLISECONDS) <= request.compute_time_limit_ms
+          )
       ) {
         val peekSource = fileStream.peek()
         val bytes = backfill.recordStrategy.calculateNextRecordBytes(peekSource)
```
shazraz, line 130: Do we need a check to make sure we dont skip once the filestream is exhausted?

jaredwindover:
Can you add some discussion of what exactly is motivating this change? I don't really understand what semantics were wrong.

mpawliszyn:
> Can you add some discussion of what exactly is motivating this change? I don't really understand what semantics were wrong.

The semantics of doing a get without an end means that you are intending on getting the whole object. What we are actually doing is a bunch of partial reads.
`Sets the optional inclusive start range within the desired object that the rest of which will be downloaded by this request.`
So if we only want part of the object we need to set some kind of end.

This appears to be causing performance problems.

mpawliszyn:
I chatted more with Jared and he thinks I am wrong. I will investigate more. In the meantime I have reached out to AWS and they recommend a retry and/or modify the client settings. You would just modify the ClientConfiguration

mpawliszyn:
AWS's answer was "just retry". Not the most satisfying.

PR #315, opened 2023-06-06
Prepare backfila for automatically publishing to maven

tso:


backfila-embedded/build.gradle.kts
```diff
@@ -18,12 +27,8 @@ dependencies {
   testImplementation(Dependencies.kotlinTest)
 }
 
-val jar by tasks.getting(Jar::class) {
-  archiveBaseName.set("backfila-embedded")
```
tso, line 22: was this `archiveBaseName.set("backfila-embedded")` doing anything?

client/build.gradle.kts
```diff
@@ -37,15 +37,16 @@ wire {
   }
 }
 
-if (rootProject.file("hooks.gradle").exists()) {
-  apply(from = rootProject.file("hooks.gradle"))
+tasks.named("kotlinSourcesJar") {
```
tso, line 40: is there a non-stringly typed way of doing this dependency?

client/build.gradle.kts
```diff
@@ -37,15 +37,16 @@ wire {
...
+  dependsOn("generateMainProtos")
 }
 
-apply(from = "$rootDir/gradle-mvn-publish.gradle")
+tasks.named("dokkaGfm") {
+  dependsOn("generateMainProtos")
+}
 
-// Prevent proto source files from conflicting
-afterEvaluate {
```
tso, line 47: doesn't seem to be needed after specifying the task dependency order!

build.gradle.kts
```diff
@@ -116,7 +117,35 @@ subprojects {
   }
 }
 
-// Disable the Gradle wrapper if Gradle is managed by Hermit
-tasks.named<Wrapper>("wrapper") {
```
tso, line 120: don't think we need this?

backfila-embedded/build.gradle.kts
swankjesse: probably only useful if you’re also using ShadowJar

client-base/build.gradle.kts
```diff
@@ -1,5 +1,8 @@
-apply(plugin = "kotlin")
-apply(plugin = "java")
+plugins {
+  kotlin("jvm")
+  `java-library`
+  id("com.vanniktech.maven.publish.base")
```
swankjesse, line 4: Everywhere I see this, I think we also need this:

```
configure<MavenPublishBaseExtension> {
  configure(
    KotlinJvm(javadocJar = Dokka("dokkaGfm"))
  )
}
```

client/build.gradle.kts
swankjesse: there’s a way that looks less stringly typed:

```
val kotlinSourcesJar by tasks.getting {
  ...
}
```

But that’s potentially worse, because the val name is special

swankjesse (approved):

client-dynamodb-v2/build.gradle.kts
```diff
@@ -43,16 +47,8 @@ dependencies {
   testImplementation(project(":client-misk"))
   // Required until DynamoDBLocal is built with antlr >4.11 which wisp-config pulls in
   testImplementation("org.antlr:antlr4-runtime:4.7.2") {
-    setForce(true)
+    version {
```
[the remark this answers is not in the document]
swankjesse: (no action to take on this, just being grumpy)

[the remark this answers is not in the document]
shellderp: read the comment above. Used by dynamodblocal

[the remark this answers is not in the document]
mpawliszyn: 🤦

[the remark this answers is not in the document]
shellderp: Why are we now specifying two different versions though? 4.7.2 and 4.9.3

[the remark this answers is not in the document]
tso: oh yeah true, dumb typo. 4.9.3 is the latest that works with dynamodb local so figured i'd bump backfila to that. will fix consistency

PR #316, opened 2023-06-07
Automatically release backfila

tso:


tso:
This requires updates to the branch settings in the repo to update the required checks — We'll need to force merge this, wait for the build to happen on master and then update the required checks

shellderp:
what is the outcome of this? it's not clear what the difference is between the old release task and the one you added

tso:
> what is the outcome of this? it's not clear what the difference is between the old release task and the one you added

Outcome is each commit will just auto publish to maven central if tests pass on master

staktrace (approved):
Didn't look at this too closely but stamping to unblock.

mpawliszyn (approved):
The tests are failing though

PR #327, opened 2023-07-26
Don't use "null" for null parameters

shellderp:


client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/hibernate/SinglePartitionHibernateTestBackfill.kt
```diff
@@ -20,7 +20,7 @@ class SinglePartitionHibernateTestBackfill @Inject constructor(
   val parametersLog = mutableListOf<SandwichParameters>()
 
   override fun backfillCriteria(config: BackfillConfig<SandwichParameters>): Query<DbMenu> {
-    return queryFactory.newQuery(MenuQuery::class).name(config.parameters.type)
+    return queryFactory.newQuery(MenuQuery::class).name(config.parameters.type ?: "chicken")
```
mpawliszyn, line 23: Maybe instead of chicken for both, null should maybe be something else like beef?

mpawliszyn (approved):

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/hibernate/SinglePartitionHibernateTestBackfill.kt
shellderp: done, that makes it clearer that it uses `chicken` when null is passed, which is kinda weird, but better than before this PR where we would pass "null"

PR #328, opened 2023-07-27
Default null not required for nullable parameters

shellderp:
Previously, you had to write
```
  data class Parameters(
    val x: String? = null,
  )
```

Now you can write

```
  data class Parameters(
    val x: String?,
  )
```


client-base/src/main/kotlin/app/cash/backfila/client/spi/BackfilaParameters.kt
```diff
@@ -86,6 +86,8 @@ class BackfilaParametersOperator<T : Any>(
             val defaultValue = defaultAnnotation.value
             map[parameter] = TYPE_CONVERTERS[parameter.type.jvmErasure]!!.invoke(defaultValue.encodeUtf8())
           }
+        } else if (!parameter.isOptional && parameter.type.isMarkedNullable) {
+          map[parameter] = null
```
shellderp, line 90: I was surprised you can set a null value and it creates an entry in the map

PR #333, opened 2023-08-08
Variants

danlu-square:
Allow registering different variants of a service, each with their own set of backfills / runs and config


client/src/main/proto/app/cash/backfila/service.proto
```diff
@@ -49,6 +49,13 @@ message ConfigureServiceRequest {
   optional string connector_extra_data = 3;
 
   optional string slack_channel = 4;
+
+  // A service can have multiple flavors, each registered independently with Backfila.
+  // Flavors have their own set of runs, use different connector configs, etc but are grouped and still identify as the same service.
+  //
+  // As an example use case, a service could register different flavors for a production and sandbox deployment.
+  // Backfila would then be able to run backfills for both deployments independently.
+  optional string flavor = 5;
```
mpawliszyn, line 58: @shellderp canadian or american spelling?

service/src/main/kotlin/app/cash/backfila/dashboard/GetServicesAction.kt
```diff
@@ -18,13 +17,13 @@ import misk.web.mediatype.MediaTypes
...
 
 class GetServicesAction @Inject constructor(
-  private val caller: @JvmSuppressWildcards ActionScoped<MiskCaller?>,
   @BackfilaDb private val transacter: Transacter,
   private val queryFactory: Query.Factory,
 ) : WebAction {
 
   data class UiService(
     val name: String,
+    val flavors: List<String?>,
```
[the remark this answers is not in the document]
mpawliszyn: Also shouldn't it be a set?

client/src/main/proto/app/cash/backfila/service.proto
mpawliszyn: Or perhaps we call it variant as @shellderp suggested?

service/src/main/kotlin/app/cash/backfila/api/ConfigureServiceAction.kt
```diff
@@ -120,5 +128,6 @@ class ConfigureServiceAction @Inject constructor(
 
   companion object {
     private val logger = getLogger<ConfigureServiceAction>()
+    const val RESERVED_FLAVOR = "flavorless"
```
danlu-square, line 131: Will change to "default", flavor -> "variant"

service/src/main/kotlin/app/cash/backfila/api/ConfigureServiceAction.kt
```diff
@@ -42,26 +42,34 @@ class ConfigureServiceAction @Inject constructor(
...
+    }
+
     val clientProvider = connectorProvider.clientProvider(request.connector_type)
     // This tests that the extra data is valid, throwing an exception if invalid.
     clientProvider.validateExtraData(request.connector_extra_data)
 
     transacter.transaction { session ->
       var dbService = queryFactory.newQuery<ServiceQuery>()
         .registryName(service)
+        .flavor(request.flavor)
```
danlu-square, line 57: Will add a limit on these

service/src/test/kotlin/app/cash/backfila/actions/CloneBackfillActionTest.kt
```diff
@@ -384,6 +389,7 @@ class CloneBackfillActionTest {
 
       val response = createBackfillAction.create(
         "deep-fryer",
+        null,
```
mpawliszyn, line 392: Can we get a test here with variant?

service/src/test/kotlin/app/cash/backfila/actions/CreateBackfillActionTest.kt
```diff
@@ -269,6 +274,7 @@ class CreateBackfillActionTest {
       assertThatThrownBy {
         createBackfillAction.create(
           "deep-fryer",
+          null,
```
mpawliszyn, line 277: we will want a variant test version

service/src/test/kotlin/app/cash/backfila/actions/CloneBackfillActionTest.kt
```diff
@@ -93,6 +93,7 @@ class CloneBackfillActionTest {
     scope.fakeCaller(user = "molly") {
       val response = createBackfillAction.create(
         "deep-fryer",
+        null,
```
danlu-square, line 96: TODO: add a basic test in these files

service/src/main/kotlin/app/cash/backfila/api/ConfigureServiceAction.kt
```diff
@@ -42,26 +42,34 @@ class ConfigureServiceAction @Inject constructor(
 
     logger.info { "Configuring service `$service` with ${request.backfills.size} backfills" }
 
+    request.flavor?.let {
+      check(!request.flavor.equals(RESERVED_FLAVOR)) { "Cannot use a reserved flavor name" }
+      check(request.flavor.isNotBlank()) { "Flavor cannot be blank" }
```
danlu-square, line 47: Any spaces will play poorly in the UI, so cover those.

service/src/test/kotlin/app/cash/backfila/actions/StartStopBackfillActionTest.kt
```diff
@@ -297,6 +303,7 @@ class StartStopBackfillActionTest {
     scope.fakeCaller(user = "molly") {
       val response = createBackfillAction.create(
         "deep-fryer",
+        null,
```
mpawliszyn, line 306: add a variant test

service/src/test/kotlin/app/cash/backfila/actions/UpdateBackfillActionTest.kt
```diff
@@ -154,6 +156,7 @@ class UpdateBackfillActionTest {
     scope.fakeCaller(user = "molly") {
       val response = createBackfillAction.create(
         "deep-fryer",
+        null,
```
mpawliszyn, line 159: We want a basic variant test.

service/src/test/kotlin/app/cash/backfila/actions/StartStopBackfillActionTest.kt
mpawliszyn, line 306: add a variant version

service/src/test/kotlin/app/cash/backfila/api/ConfigureServiceActionTest.kt
```diff
@@ -174,6 +175,117 @@ class ConfigureServiceActionTest {
...
+      assertThat(backfillNames("deep-fryer")).containsExactly("xyz")
+      assertThat(deletedBackfillNames("deep-fryer")).containsExactly("zzz")
+
+      assertThat(backfillNames("deep-fryer", "deep-fried")).containsExactly("zzz")
+      assertThat(deletedBackfillNames("deep-fryer", "deep-fried")).isEmpty()
+    }
+  }
+
+  @Test
+  fun configureMultipleFlavors_updateASpecificFlavor() {
```
mpawliszyn, line 244: use backtick!

service/src/main/kotlin/app/cash/backfila/api/ConfigureServiceAction.kt
danlu-square: Added a limit of 10

service/src/main/resources/migrations/v016__backfila.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE services
+  ADD COLUMN variant VARBINARY(100) NULL DEFAULT NULL,
+  DROP KEY `unq_registry_name`,
+  ADD UNIQUE KEY `unq_registry_name_variant` (registry_name, variant);
```
shellderp, line 4: when this is null the unique key does nothing

service/src/test/kotlin/app/cash/backfila/actions/GetServiceVariantsActionTest.kt
```diff
@@ -0,0 +1,149 @@
...
+      configureServiceAction.configureService(
+        ConfigureServiceRequest.Builder()
+          .variant("under-baked")
+          .connector_type(Connectors.ENVOY)
+          .build(),
+      )
+    }
+
+    scope.fakeCaller(user = "molly") {
+      Assertions.assertThat(getServiceVariantsAction.variants("deep-fryer").variants).containsOnly(
```
shellderp, line 141: import method

service/web/tabs/app/src/components/ServicesListComponent.tsx
```diff
@@ -3,6 +3,26 @@ import { ErrorCalloutComponent } from "@misk/core"
...
+function ServiceLink(props: IServiceLinkProps) {
+  if (props.variants.length > 1) {
+    return (
+      <Link to={`/app/services/${props.name}/variants/`}>
+        {this.props.name}
+      </Link>
+    );
+  }
+  return (
+    <Link to={`/app/services/${props.name}/variants/${props.variants[0]}`}>
```
shellderp, line 20: if there's only one variant, I wouldn't change anything about how the UI works. This looks like it will break old urls.

service/web/tabs/app/src/routes/index.tsx
```diff
@@ -7,21 +7,23 @@ import {
...
+  ServiceVariantsContainer,
   LayoutContainer
 } from "../containers"
 
 const routes = (
   <div>
     <Switch>
       <Route
-        path="/app/services/:service/create"
+        path="/app/services/:service/variants/:variant/create"
```
shellderp, line 18: breaking change

client/src/main/proto/app/cash/backfila/service.proto
```diff
@@ -76,6 +83,7 @@ message CreateBackfillResponse {
 // Creates a backfill under the calling service and starts it.
 message CreateAndStartBackfillRequest {
   optional CreateBackfillRequest create_request = 1;
+  optional string variant = 2;
```
shellderp, line 86: why is this not part of `CreateBackfillRequest` ?

service/src/main/kotlin/app/cash/backfila/api/ConfigureServiceAction.kt
```diff
@@ -32,6 +32,8 @@ class ConfigureServiceAction @Inject constructor(
   private val clock: Clock,
   private val connectorProvider: ConnectorProvider,
 ) : WebAction {
+  private val WHITESPACE_REGEX = Regex("\\s")
```
shellderp, line 35: move to companion object

client/src/main/proto/app/cash/backfila/service.proto
```diff
@@ -49,6 +49,13 @@ message ConfigureServiceRequest {
   optional string connector_extra_data = 3;
 
   optional string slack_channel = 4;
+
+  // A service can have multiple variants, each registered independently with Backfila.
+  // Variants have their own set of runs, use different connector configs, etc but are grouped and still identify as the same service.
+  //
+  // As an example use case, a service could register different variants for a production and sandbox deployment.
```
shellderp, line 56: question, why not run a separate backfila instance in the sandbox deployment and register all the services there if this is your goal?

service/src/main/resources/migrations/v016__backfila.sql
danlu-square: Good call. Do you have a preference for how to approach this?

Off the top of my head, we could:
1. Require a backfill of the column with a default value on the next Backfila upgrade
2. Workaround using [generated columns](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html). This places a requirement that Backfila must be on MySQL 5.7.
3. Take a lock at the top of the transaction and rely on the logic in code to enforce the constraint.

Something else?

client/src/main/proto/app/cash/backfila/service.proto
danlu-square: It could be. I put it here to start with because the fields in `CreateBackfillRequest` seemed to moreso be properties of the run, rather than the service + variant that registered it.

service/src/main/resources/migrations/v016__backfila.sql
shellderp: If you want to make "default" the default value, I think you can specify the default value when adding the column and not require a backfill

danlu-square: Oh, good point. That seems easiest

danlu-square: Ok, this is non-nullable with a default value now

service/web/tabs/app/src/components/ServicesListComponent.tsx
danlu-square: As a first pass, I've left this piece of code as-is.

Behaviorally, this should be the same as the old UI. If there's a single variant, the UI will proceed directly to the original page. If there are multiple, it takes the user to a subpage listing out variants.

For old urls, I've added redirects that will send the user to the new page, using `RESERVED_VARIANT` for the dynamic segment.

client/src/main/proto/app/cash/backfila/service.proto
mpawliszyn: It will need to be part of both I think?

client/src/main/proto/app/cash/backfila/service.proto
mpawliszyn: Yeah my sense is that this problem will get worse here over time. That people will want different types beyond just sandbox.

service/src/main/kotlin/app/cash/backfila/dashboard/CreateBackfillAction.kt
```diff
@@ -27,10 +28,11 @@ class CreateBackfillAction @Inject constructor(
   @Authenticated(capabilities = ["users"])
   fun create(
     @PathParam service: String,
+    @QueryParam variant: String? = null,
```
mpawliszyn, line 31: should this be default from that companion object?

mpawliszyn: RESERVED_VARIANT?

service/web/tabs/app/src/containers/ServiceContainer.tsx
```diff
@@ -17,12 +17,13 @@ class ServiceContainer extends React.Component<
...
   private service: string = (this.props as any).match.params.service
-  private backfillRunsTag: string = `${this.service}::BackfillRuns`
+  private variant: string = (this.props as any).match.params.variant
+  private backfillRunsTag: string = `${this.service}::${this.variant}::BackfillRuns`
 
   componentDidMount() {
     this.props.simpleNetworkGet(
       this.backfillRunsTag,
-      `/services/${this.service}/backfill-runs`
+      `/services/${this.service}/backfill-runs?variant=${this.variant}`
```
mpawliszyn, line 26: The mixing in the ways of specifying the url seems odd. We should be consistent.

client/src/main/proto/app/cash/backfila/service.proto
mpawliszyn: It won't as the CreateBackfillRequest is used by the UI rather than the App

mpawliszyn:
I pushed a change for the service side comments work. I/We still need to fix the web part as well. I think we might need a few more Components or something but I am not an expert so this might take a bit of time.

service/web/tabs/app/src/routes/index.tsx
mpawliszyn: Only adding now.

service/src/main/kotlin/app/cash/backfila/dashboard/CreateBackfillAction.kt
```diff
@@ -25,12 +26,33 @@ class CreateBackfillAction @Inject constructor(
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
   // TODO allow any user
   @Authenticated(capabilities = ["users"])
+  fun createDefault(
```
mpawliszyn, line 29: On second thought I think I should just remove these as these will be hit by the javascript and therefore will never be used. @shellderp happy to discuss more if there is a case I am forgetting but I think I should remove these.

mpawliszyn (approved):
I believe we have addressed all the comments now.

PR #340, opened 2023-10-16
Extend management client to return ID on start and check status

abhijitsverma:
Enable job orchestration by extending the management client to 
(1) Return a run ID from the createAndStart api
(2) Provide an api for check status for a job already started


client/src/main/kotlin/app/cash/backfila/client/BackfilaManagementClient.kt
```diff
@@ -14,5 +16,9 @@ interface BackfilaManagementClient {
     scan_size: Long? = null,
     pkey_range_start: ByteString? = null,
     pkey_range_end: ByteString? = null,
-  )
+  ): CreateAndStartResponse?
```
mpawliszyn, line 19: This is just for Java and Kotlin. We don't want protos. There is no actual networking client involved.

client/src/main/kotlin/app/cash/backfila/client/BackfilaManagementClient.kt
```diff
@@ -14,5 +16,9 @@ interface BackfilaManagementClient {
...
+
+  fun checkStatus(
+    backfill_run_id: Long,
```
mpawliszyn, line 22: I am really having second thoughts about this. If you need strong control over the execution and status I don't think Backfila is the solution you are looking for.

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfilaManagementClient.kt
```diff
@@ -44,5 +47,21 @@ class RealBackfilaManagementClient @Inject internal constructor(
...
         ).build(),
     )
+    return CreateAndStartResponse.Builder().backfill_run_id(createResponse.backfill_run_id).build()
+  }
+
+  override fun checkStatus(backfill_run_id: Long): CheckStatusResponse? {
+    val statusResponse = client.checkBackfillStatus(
+      CheckBackfillStatusRequest.Builder().backfill_run_id(backfill_run_id).build(),
+    )
+    val internalStatus = statusResponse.status.let {
```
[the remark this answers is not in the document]
mpawliszyn: The question left here is do we want to return the Status proto [here]( https://github.com/cashapp/backfila/blob/master/client/src/main/proto/app/cash/backfila/service.proto#L93) or if we want to return a string?

The case for returning a string is that we may need to change some of the semantics or add new states so a layer of indirection will make that more obvious. @shellderp any opinion?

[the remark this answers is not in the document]
abhijitsverma: String sounds fine

client/src/main/kotlin/app/cash/backfila/client/BackfilaManagementClient.kt
abhijitsverma: Happy to remove things that are not needed. Wrote the protos in as a generally followed pattern; can be removed. The thing thats most relevant is being able to know which run ID was triggered which is a current gap. Not really looking for stronger control over anything on the backfila side

abhijitsverma: The checkStatus is just a nice to have to be able to know if the job is running; no expectation change on how it works :)

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfilaManagementClient.kt
```diff
@@ -22,8 +23,8 @@ class RealBackfilaManagementClient @Inject internal constructor(
     scan_size: Long?,
     pkey_range_start: ByteString?,
     pkey_range_end: ByteString?,
-  ) {
-    client.createAndStartBackfill(
+  ): Long? {
```
mpawliszyn, line 26: I don't think it can be null

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfilaManagementClient.kt
```diff
@@ -44,5 +45,14 @@ class RealBackfilaManagementClient @Inject internal constructor(
             .build(),
         ).build(),
     )
+    return createResponse.backfill_run_id
+  }
+
+  override fun checkStatus(backfill_run_id: Long): String? {
```
mpawliszyn, line 51: This string shouldn't be null either.

client/src/main/kotlin/app/cash/backfila/client/BackfilaManagementClient.kt
```diff
@@ -14,5 +14,9 @@ interface BackfilaManagementClient {
     scan_size: Long? = null,
     pkey_range_start: ByteString? = null,
     pkey_range_end: ByteString? = null,
-  )
+  ): Long?
```
mpawliszyn, line 17: Yeah remove these nullable return values from this interface.

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/hibernate/BackfilaManagementClientTest.kt
```diff
@@ -35,12 +35,22 @@ class BackfilaManagementClientTest {
...
+    assertThat(run_id).isNotNull()
+  }
+
+  @Test
+  fun `check status`() {
+    val run_id =
+      managementClient.createAndStart(ChickenToBeefBackfill::class.java, dry_run = false)
+
+    val status = managementClient.checkStatus(run_id!!)
+    assertThat(status).isEqualTo("COMPLETE")
```
mpawliszyn, line 53: We probably want one for Running too.

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfilaManagementClient.kt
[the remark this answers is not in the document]
shellderp: I sorta prefer proto actually, it is very rare we will add new states but having the set of possible values defined is nice.

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/hibernate/BackfilaManagementClientTest.kt
```diff
@@ -35,12 +35,22 @@ class BackfilaManagementClientTest {
       session.save(DbMenu("chicken"))
     }
 
-    managementClient.createAndStart(ChickenToBeefBackfill::class.java, dry_run = false)
+    val runId = managementClient.createAndStart(ChickenToBeefBackfill::class.java, dry_run = false)
```
mpawliszyn, line 38: Can you add a few tests for this in `client-base`?

mpawliszyn (commented):
I am working on an answer around the proto vs the string question.

Once those tests are added and I have figured the above out I think it is ready to merge.

mpawliszyn:
Both @swankjesse and @shellderp are team proto enum, so let's convert that back. Sorry for the churn.

PR #341, opened 2023-10-27
Update the s3 client to handle that the end range is inclusive for Amazon S3.

mpawliszyn:


client-s3/src/main/kotlin/app/cash/backfila/client/s3/shim/FakeS3Service.kt
```diff
@@ -55,7 +55,15 @@ class FakeS3Service @Inject constructor() : S3Service {
     key: String,
     seekStart: Long,
     seekEnd: Long,
-  ): ByteString = files[bucket to key]?.substring(seekStart.toInt(), seekEnd.toInt()) ?: EMPTY
+  ): ByteString {
+    /*
+      S3's `withRange` has an inclusive end, so replicate that behaviour here.
+      This will also fail if bytes are requested after the file end which Amazon does not do. However,
+      this holds us to a higher standard to hopefully discover unintended behaviour.
```
shellderp, line 62: +1

shellderp (approved):

PR #343, opened 2023-11-09
Add API to respond with for running backfills

jrobotham-square:
When an incident occurs it's often really valuable to be able to see what Backfills are currently running in order to:
- Identify ones that may be causing or exacerbating issues.
- Identify ones that could be stopped to reduce the risk.

This PR exposes a new endpoint to surface the currently running backfills.

As part of this change I've also included instructions on starting mysql locally for development and fixed some issues that occur locally when building / running tests on Apple silicon.

Subsequent PR will integrate this endpoint into the UI to create a screen something like:
Screenshot 2023-11-09 at 8 37 16 am

### DIFF
--- a/service/src/test/kotlin/app/cash/backfila/actions/GetActiveBackfillsActionTest.kt
+++ b/service/src/test/kotlin/app/cash/backfila/actions/GetActiveBackfillsActionTest.kt
@@ -132,6 +132,7 @@ class GetActiveBackfillsActionTest {
     scope.fakeCaller(user = "molly") {
       val response = createBackfillAction.create(
         serviceName,
+        "variant1",
         CreateBackfillRequest.Builder()
           .backfill_name(backfillName)
           .dry_run(dryRun)
@@ -151,6 +152,7 @@ class GetActiveBackfillsActionTest {
         ConfigureServiceRequest.Builder()
           .backfills(backfills)
           .connector_type(Connectors.ENVOY)
+          .variant("variant1")
           .build(),
       )
     }


adrw:
Mind reposting the screenshots, they don't seem to be showing in the preview?

mpawliszyn:
Make sure to rebase on top of the variant PR.

service/src/main/kotlin/app/cash/backfila/dashboard/GetActiveBackfillsAction.kt
```diff
@@ -0,0 +1,64 @@
...
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+
+class GetActiveBackfillsAction @Inject constructor(
+  private val caller: @JvmSuppressWildcards ActionScoped<MiskCaller?>,
+  @BackfilaDb private val transacter: Transacter,
+  private val queryFactory: Query.Factory,
+) : WebAction {
+
+  @Get("/backfills/running")
```
mpawliszyn, line 26: Probably needs to be paginated.

jrobotham-square:
> Mind reposting the screenshots, they don't seem to be showing in the preview?

👍  done.  (not sure what happened there)

service/src/main/kotlin/app/cash/backfila/dashboard/GetActiveBackfillsAction.kt
jrobotham-square: yeah - was kinda assuming the number of backfills concurrently running wouldn't generally be enough to require pagination, but that's I guess that's not necessarily the case.

jrobotham-square: Actually, looks like we don't have pagination on any of the endpoints apart from the `/services/{service}/variants/{variant}/backfill-runs` one.  (which makes sense because it's completely unbounded)

Feels like we might be able to avoid it for this endpoint given it's only returning currently running ones.  WDYT?

jrobotham-square:
> Make sure to rebase on top of the variant PR.

I've rebased.  It's not immediately obvious to me how the variant changes work, I've just hardcoded some values to make the tests I added pass.

service/src/main/kotlin/app/cash/backfila/dashboard/GetActiveBackfillsAction.kt
mpawliszyn: https://github.com/cashapp/backfila/blob/master/service/src/main/kotlin/app/cash/backfila/dashboard/GetBackfillRunsAction.kt#L59

mpawliszyn: You are listing runs so of course it needs to be paginated. 

The other ones might need a rethink soon given how popular Backfila has become.

Why did you pick runs to show rather than services?

PR #346, opened 2023-11-13
Rearranging the routing rules so variant components are used.

mpawliszyn:
While testing we found that the router isn't making it to the variant pages so I have moved them.

### DIFF
--- a/service/web/tabs/app/src/routes/index.tsx
+++ b/service/web/tabs/app/src/routes/index.tsx
@@ -1,6 +1,5 @@
 import * as React from "react"
 import { Route, Switch } from "react-router"
-import { Redirect } from "react-router-dom"
 import {
   BackfillStatusContainer,
   CloneFormContainer,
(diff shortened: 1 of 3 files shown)


adrw (approved):

service/web/tabs/app/src/routes/index.tsx
```diff
@@ -11,45 +10,40 @@ import {
...
-        path="/app/services/:service"
-        component={ServiceDetailsContainer}
+        path="/app/services/:service/runs/:offset"
+        component={ServiceRunsContainer}
       />
       <Route
-        path="/app/services/:service/variants/:variant"
-        component={ServiceDetailsContainer}
-      />
-      <Redirect
```
adrw, line 46: This no longer should be a redirect?

adrw (approved):

service/web/tabs/app/src/routes/index.tsx
mpawliszyn: correct. This shouldn't have been there.

PR #349, opened 2023-11-15
Fix a couple warnings and remove the slash from the variants get call.

mpawliszyn:


keeferrourke (approved):

mpawliszyn:
I did a blob copy paste but @adrw and I will be moving UI frameworks soon anyway. So we will fix that when we do the move.

adrw (approved):

PR #350, opened 2023-11-20
Allow service list to be filtered by one that have running backfills

jrobotham-square:
As a short term / quick fix for visualising backfills that are currently running, this adds the ability to filter the service list so we only show ones where running backfills is > 0.

- Add toggle to switch between modes.
- Use query parameter to allow linking directly

https://github.com/cashapp/backfila/assets/92703335/e9664f47-bad4-4ace-bba3-2a079fa3ab6f

### DIFF
--- a/service/web/tabs/app/src/containers/HomeContainer.tsx
+++ b/service/web/tabs/app/src/containers/HomeContainer.tsx
@@ -8,30 +8,55 @@ import {
   mapDispatchToProps,
   mapStateToProps
 } from "src/ducks"
+import {Checkbox, FormGroup} from "@blueprintjs/core"
 import { LayoutContainer } from "."
+import { RouteComponentProps } from 'react-router-dom';
 
-class TabContainer extends React.Component<IState & IDispatchProps, IState> {
+
+
+interface TabContainerState extends RouteComponentProps {
+  only_show_running_backfills: boolean
+}
+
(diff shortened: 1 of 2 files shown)


service/web/tabs/app/src/containers/HomeContainer.tsx
```diff
@@ -8,30 +8,55 @@ import {
...
-    )
+        if (!this.state) {
+          return (<div>loading...</div>)
+        } else {
+          return (
+            <LayoutContainer>
+              <FormGroup>
+                <Checkbox checked={this.state.only_show_running_backfills}
+                                label={"Only Services with running Backfills"}
+                                disabled={false}
```
davidcarlin-ap, line 42: Can we have it defaulted to checked?
```suggestion
                                disabled={false}
                                defaultChecked={true}
```

jrobotham-square: I'd be happy to change it.   

The only reason I left it unchecked by default was not wanting to interrupt people that might use the index as a starting point to jump to their service.

adrw (approved):
Nicely done!

mpawliszyn (approved):

Rizziepit:
I just noticed the existence of this tickbox in the UI and wanted to give @jrobotham-square a 🥇 for that. Thank you!!

PR #339, opened 2023-10-16
Move DevelopmentService to test to allow for logging to stdout by default.

mpawliszyn:

PR #351, opened 2023-11-20
Create an endpoint to see some data into the DB to make UI testing easier

jrobotham-square:
Creates an endpoint in the dev server that you can hit to run some code to populated a bunch of services and backfills into the database.

(just a possible approach we could take)

### DIFF
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -67,6 +67,7 @@ dependencies {
   implementation(Dependencies.wispConfig)
   implementation(Dependencies.wispDeployment)
   implementation(Dependencies.wispLogging)
+  implementation(Dependencies.slf4jSimple)
 
   implementation(project(":client"))
 
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
@@ -1,0 +1,77 @@
+package app.cash.backfila.dashboard
+
+import app.cash.backfila.api.ConfigureServiceAction
+import app.cash.backfila.api.CreateAndStartBackfillAction
+import app.cash.backfila.protos.service.ConfigureServiceRequest
+import app.cash.backfila.protos.service.CreateAndStartBackfillRequest
+import app.cash.backfila.protos.service.CreateBackfillRequest
+import javax.inject.Inject
+import misk.MiskCaller
+import misk.inject.keyOf
+import misk.scope.ActionScope
+import misk.web.Get
+import misk.web.ResponseContentType
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+
+class SeedDbAction @Inject constructor(
+  val configureServiceAction: ConfigureServiceAction,
+  val createAndStartBackfillAction: CreateAndStartBackfillAction,
+  val scope: ActionScope,
+) : WebAction {
+
+  @Get("/seed")
+  @ResponseContentType(MediaTypes.APPLICATION_JSON)
+  fun seed(): String {
+    scope.close()
+    createServiceWithBackfills("deep-fryer", listOf("ChickenSandwich"))
+    startBackfill("deep-fryer", "ChickenSandwich")
+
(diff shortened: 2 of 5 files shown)


service/build.gradle.kts
```diff
@@ -67,6 +67,7 @@ dependencies {
   implementation(Dependencies.wispConfig)
   implementation(Dependencies.wispDeployment)
   implementation(Dependencies.wispLogging)
+  implementation(Dependencies.slf4jSimple)
```
jrobotham-square, line 70: Not sure if this is a good idea.  Looks like something similar is happening here:
https://github.com/cashapp/backfila/pull/339

adrw (approved):
Huge fan on this approach, would help my UI work a lot too!

service/build.gradle.kts
mpawliszyn: This might be why the tests are failing.

service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
```diff
@@ -0,0 +1,77 @@
+package app.cash.backfila.dashboard
```
mpawliszyn, line 1: I'm fine with this in the test or development app. Not so much with the general service though. Maybe just break DevelopmentApp out into another module?

service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
```diff
@@ -0,0 +1,77 @@
...
+      createServiceWithBackfills("deep-fryer-$i", listOf("ChickenSandwich"))
+    }
+
+    return "yo"
+  }
+
+  private fun startBackfill(serviceName: String, backfillName: String) {
+    scope.fakeCaller(service = serviceName) {
+      val response = createAndStartBackfillAction.createAndStartBackfill(
+        CreateAndStartBackfillRequest.Builder()
```
mpawliszyn, line 40: Does this work? The backfill itself it likely failing? What are you trying to test with this?

service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
```diff
@@ -0,0 +1,77 @@
...
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+
+class SeedDbAction @Inject constructor(
+  val configureServiceAction: ConfigureServiceAction,
+  val createAndStartBackfillAction: CreateAndStartBackfillAction,
+  val scope: ActionScope,
+) : WebAction {
+
+  @Get("/seed")
```
mpawliszyn, line 23: Rather than an endpoint don't we want a startup action?

service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
mpawliszyn: I wonder if we just want to start some other test only development services. And those could have these kinds of apis.

mpawliszyn (commented):
This has given me the idea of having at least 2 test only dev apps That we can start up and use. I think that is where the create and start apis and things should go.

service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
jrobotham-square: I think that's actually the right way to do it, but didn't know offhand how/where to hook that in.

service/src/main/kotlin/app/cash/backfila/dashboard/SeedDbAction.kt
jrobotham-square: My testing was just UI testing filtering by services that have running backfills and this seems to end up with the backfills sitting in "Running" so worked for my testing.   (but probably won't be useful for other more indepth testing)

adrw:
Closing since the development services were merged in https://github.com/cashapp/backfila/pull/355

PR #355, opened 2023-11-23
Adds a fake client development service FineDining.

mpawliszyn:
This way you can start up Backfila and FineDining to showcase actual Backfila behaviour.

### DIFF
--- a/service/src/test/kotlin/app/cash/backfila/development/ServiceHeaderInterceptor.kt
+++ b/service/src/test/kotlin/app/cash/backfila/development/ServiceHeaderInterceptor.kt
@@ -1,0 +1,33 @@
+package app.cash.backfila.development
+
+import app.cash.backfila.client.BackfilaApi
+import jakarta.inject.Inject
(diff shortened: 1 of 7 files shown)


service/src/test/kotlin/app/cash/backfila/development/ServiceHeaderInterceptor.kt
```diff
@@ -0,0 +1,33 @@
...
+import jakarta.inject.Singleton
+import okhttp3.Interceptor
+import okhttp3.Response
+import retrofit2.Invocation
+
+/**
+ * Allows Development services to provide Backfila an authorized name.
+ */
+@Singleton
+class ServiceHeaderInterceptor @Inject constructor(
```
adrw, line 14: This is so dope, we probably should upstream this to Misk in the same jar as FakeCallerAuthenticator.

adrw (approved):

shellderp:
cool!

PR #356, opened 2023-11-24
Adding another fake Development service McDees.

mpawliszyn:
Boots and Cats and...


service/src/test/kotlin/app/cash/backfila/development/BackfilaDevelopmentService.kt
```diff
@@ -131,3 +132,12 @@ internal class DevelopmentViewLogsUrlProvider : ViewLogsUrlProvider {
     return "/"
   }
 }
+
+class DevServiceConstants {
```
adrw, line 136: Niceee

adrw (approved):
Super hype, thanks for doing this!

PR #366, opened 2024-01-10
Make Utf8StringNewlineStrategy class open

elmerland:
Make Utf8StringNewlineStrategy class open so that clients can extend the class and apply custom transformations.

For example:
```kotlin
class MyRecordStrategy() : Utf8StringNewlineStrategy(ignoreBlankLines = true) {
  override fun calculateNextRecordBytes =
    super.calculateNextRecordBytes(source).map(String::uppercase)
}

```

### DIFF
--- a/client-s3/src/main/kotlin/app/cash/backfila/client/s3/record/Utf8StringNewlineStrategy.kt
+++ b/client-s3/src/main/kotlin/app/cash/backfila/client/s3/record/Utf8StringNewlineStrategy.kt
@@ -6,7 +6,7 @@ import okio.ByteString
 /**
  * Produces single string records separated by newlines.
  */
-class Utf8StringNewlineStrategy(
+open class Utf8StringNewlineStrategy(
   private val ignoreBlankLines: Boolean = true,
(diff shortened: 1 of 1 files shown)


adrw (approved):

mpawliszyn (approved):

client-s3/src/main/kotlin/app/cash/backfila/client/s3/record/Utf8StringNewlineStrategy.kt
```diff
@@ -6,7 +6,7 @@ import okio.ByteString
 /**
  * Produces single string records separated by newlines.
  */
-class Utf8StringNewlineStrategy(
+open class Utf8StringNewlineStrategy(
```
shellderp, line 9: wouldn't composition work well for this instead of inheritance?

elmerland: it would yes! I kept the changes small because I didn't have a lot of time.

PR #368, opened 2024-01-26
Add stop all backfills action

mhickman:
This is meant to be used as an emergency kill switch to stop all running backfills during some type of incident.

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/StopAllBackfillsAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/StopAllBackfillsAction.kt
@@ -1,0 +1,59 @@
+package app.cash.backfila.dashboard
+
+import app.cash.backfila.service.persistence.BackfilaDb
+import app.cash.backfila.service.persistence.BackfillRunQuery
+import app.cash.backfila.service.persistence.BackfillState
+import javax.inject.Inject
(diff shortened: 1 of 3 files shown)


Patrax (approved):

adrw (approved):

service/src/main/kotlin/app/cash/backfila/dashboard/StopAllBackfillsAction.kt
```diff
@@ -0,0 +1,59 @@
...
+  @ResponseContentType(MediaTypes.APPLICATION_JSON)
+  @Authenticated(capabilities = ["users"])
+  fun stopAll(
+    @Suppress("unused") @RequestBody
+    request: StopAllBackfillsRequest,
+  ): StopAllBackfillsResponse {
+    // TODO check user has permissions for this service with access api
+    logger.info { "Stop all backfills called by ${caller.get()?.user}" }
+
+    val runningBackfills = transacter.transaction { session ->
```
shellderp, line 42: it's not a good idea to return Db objects out of the transaction, can you map this to return their ids?

mhickman: sure thing https://github.com/cashapp/backfila/pull/370

PR #372, opened 2024-01-29
Add emergency text box input to enable "stop all backfills" button.

mhickman:
Adds a small bit of friction to reduce chances someone hits this button
on accident.

Kapture 2024-01-29 at 10 17 05

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/StopAllBackfillsAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/StopAllBackfillsAction.kt
@@ -47,7 +47,7 @@ class StopAllBackfillsAction @Inject constructor(
     }
 
     runningBackfillIds.forEach { backfillId ->
-      logger.info { "Stop backfill ${backfillId} by ${caller.get()?.user}" }
+      logger.info { "Stop backfill $backfillId by ${caller.get()?.user}" }
       backfillStateToggler.toggleRunningState(backfillId, caller.get()!!, BackfillState.PAUSED)
     }
 
--- a/service/web/tabs/app/src/components/StopAllBackfillsButton.tsx
+++ b/service/web/tabs/app/src/components/StopAllBackfillsButton.tsx
@@ -1,23 +1,29 @@
 import * as React from "react"
 import Axios from "axios"
-import { Button, Intent, Toaster, Classes, Popover } from "@blueprintjs/core"
+import { Button, Intent, Toaster, Classes, Popover, InputGroup } from "@blueprintjs/core"
 
-interface IStopAllBackillsProps {
-  onUpdate?: () => void
+interface IStopAllRealBackillsProps {
+  onUpdate?: () => void,
+  disabled: boolean
 }
 
-interface IStopAllBackfillsState {
+interface IStopAllRealBackfillsState {
   loading: boolean
 }
 
-class StopAllRealButton extends React.Component<IStopAllBackillsProps, IStopAllBackfillsState> {
-  public state: IStopAllBackfillsState = {
+const EMERGENCY_PROMISE = "I realize that by breaking glass that I am imposing a cost on the whole organization"
+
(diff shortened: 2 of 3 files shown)


service/src/main/kotlin/app/cash/backfila/dashboard/StopAllBackfillsAction.kt
```diff
@@ -47,7 +47,7 @@ class StopAllBackfillsAction @Inject constructor(
     }
 
     runningBackfillIds.forEach { backfillId ->
-      logger.info { "Stop backfill ${backfillId} by ${caller.get()?.user}" }
+      logger.info { "Stop backfill $backfillId by ${caller.get()?.user}" }
       backfillStateToggler.toggleRunningState(backfillId, caller.get()!!, BackfillState.PAUSED)
     }
```
mpawliszyn, line 52: I think I also want to make sure that slack messaging is tested when we have an action like this. Also I think we might want to change the slack message a bit. But that isn't part of this PR.

service/web/tabs/app/src/components/StopAllBackfillsButton.tsx
```diff
@@ -67,8 +99,8 @@ class StopAllButton extends React.Component {
...
         }
       />
-    );
+    )
   }
 }
 
-export default StopAllButton
\ No newline at end of file
+export default StopAllButton
```
mpawliszyn, line 106: @adrw where does this fit in our V1 vs V2 UI? Where are we on that? Maybe we should push on finishing that?

service/web/tabs/app/src/components/StopAllBackfillsButton.tsx
```diff
@@ -46,7 +51,24 @@ class StopAllRealButton extends React.Component<IStopAllBackillsProps, IStopAllB
...
+}
+
+interface IStopAllBackfillsState {
+  disabled: boolean
+}
+
+class StopAllButton extends React.Component<IStopAllBackillsProps, IStopAllBackfillsState> {
+  handleTextChange(newValue: String) {
+    this.setState({
+      disabled: newValue != "break glass"
```
mpawliszyn, line 64: `I realize that by breaking glass that I am imposing a cost on the whole Block organization.` ?

service/web/tabs/app/src/components/StopAllBackfillsButton.tsx
```diff
@@ -55,8 +77,18 @@ class StopAllButton extends React.Component {
         position="right"
         content={
           <div>
-            <p>Are you sure? This should only been done during large incidents.</p>
-            <StopAllRealButton />
+            <p>
+              Are you sure? This should only been done during large incidents.
```
mpawliszyn, line 81: The right place for an idiot box.

mpawliszyn (approved):

service/web/tabs/app/src/components/StopAllBackfillsButton.tsx
mpawliszyn: Since this is a public repo though:
`I realize that by breaking glass that I am imposing a cost on the whole organization.`

shellderp: bit too condescending imo

mpawliszyn: Which part do you have issue with?
`break glass and stop ALL backfills` ?

mhickman: yeah sure

mhickman: didn't see @shellderp 's comments originally, I think github was havin a day - happy to make this into whatever folks it to say

mdboop (commented):
Late to the party here, but thank you for adding this so quickly.

PR #377, opened 2024-02-23
Renaming ClientServiceClient to CallbackConnector for clarity.

mpawliszyn:
This also avoids a name collision when trying to add the grpc callback option.


adrw (approved):
Nice, I like callback as a name. Not sure about client vs connector, don't feel strongly either way. CallbackClient could be a good one.

ThePumpingLemma (approved):

PR #379, opened 2024-03-19
WIP: specify total backfilled records is in bytes

danieloh0714:
The UI for a backfill displays the total backfilled records as well as total records to run. 

image

This is a bit misleading, because although this sounds like it is referring to the number of database items/records, this actually signifies bytes. This PR makes that clear.


adrw:
Closing in favor of https://github.com/cashapp/backfila/pull/380

PR #326, opened 2023-07-24
Adds dynamodb sdk v1 backfill on index

jsibbison-square:
Allows the backfill to specify an indexName to be used in the scan rather than the main table.

PR #382, opened 2024-04-15
Adds a way to backfill dynamo gsi's for the v2 sdk client.

mpawliszyn:
This is the V2 version of https://github.com/cashapp/backfila/pull/326

### DIFF
--- a/client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
+++ b/client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
@@ -31,7 +31,7 @@ abstract class UpdateInPlaceDynamoDbBackfill<I : Any, P : Any>(
             dynamoDbTable.tableName() to itemsToSave.map {
               WriteRequest.builder().putRequest(
                 PutRequest.builder().item(
-                  dynamoDbTable.tableSchema().itemToMap(it, false),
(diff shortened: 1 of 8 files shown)


client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
```diff
@@ -31,7 +31,7 @@ abstract class UpdateInPlaceDynamoDbBackfill<I : Any, P : Any>(
             dynamoDbTable.tableName() to itemsToSave.map {
               WriteRequest.builder().putRequest(
                 PutRequest.builder().item(
-                  dynamoDbTable.tableSchema().itemToMap(it, false),
+                  dynamoDbTable.tableSchema().itemToMap(it, true),
```
adrw, line 34: What does the true/false toggle here?

adrw (approved):

client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
mpawliszyn: If someone ever requires this. Then that means they would know how to configure Dynamo to handle this properly. I couldn't figure this out from reading the docs and searching for answers.

I think this may be a forward looking feature rather than what the current state can accept.

PR #386, opened 2024-05-13
Adding annotation for java null parameter defaults and more checks.

mpawliszyn:


client-base/src/main/kotlin/app/cash/backfila/client/internal/BackfilaStartupConfigurator.kt
```diff
@@ -34,38 +34,49 @@ class BackfilaStartupConfigurator @Inject constructor(
...
-
-    // TODO(mikepaw): make this async.
-    try {
-      backfilaClient.configureService(request)
-
-      logger.info {
-        "Backfila lifecycle listener initialized. " +
-          "Updated backfila with ${registrations.size} backfills."
-      }
+    val request: ConfigureServiceRequest? = try {
```
keeferrourke, line 37: nit: this try/catch block might be a little easier to read if it was extracted to its own method.

keeferrourke: Maybe something like:

```suggestion
    val request: ConfigureServiceRequest? = if (backfilaClient.throwOnStartup) {
      getRequest().getOrThrow()
    } else {
      getRequest().getOrNull()
    }
    if (request == null) {
       logger.error(e) { "Exception generating backfila registration request, skipped!" }
    }
```

where:

```kt
private fun getRequest(): Result<ConfigureServiceRequest> = runCatching {
  ConfigureServiceRequest.Builder()...
}
```

client-base/src/test/java/app/cash/backfila/embedded/CaseParameters.java
```diff
@@ -9,14 +10,21 @@ public class CaseParameters {
...
+  public final String required;
 
   public CaseParameters(
       @Description(text = "Whether to change case to upper case or lower case.")
       @BackfilaDefault(name = "casing", value = "upper") String casing,
       @BackfilaDefault(name = "testLong", value = "123") long testLong,
       @BackfilaDefault(name = "testInt", value = "789") int testInt,
       @BackfilaDefault(name = "testBool", value = "false") boolean testBool,
-      @BackfilaRequired(name = "required") String requiredParameter) {
+      @BackfilaNullDefault(name = "testNullString") String testNullString,
```
keeferrourke: Should we enforce `@Nullable` on these types of parameters?

keeferrourke: Or rather... instead of introducing a new annotation, can we implement the same behaviour when a parameter is annotated like `@BackfilaDefault(name = "foo") @Nullable String foo`?

Then we complain if no value is set, and the parameter isn't annotated as nullable?

keeferrourke (approved):

client-base/src/test/java/app/cash/backfila/embedded/CaseParameters.java
keeferrourke: Never mind I guess I was reviewing an old diff!

shellderp (approved):

client-base/src/main/kotlin/app/cash/backfila/client/internal/BackfilaStartupConfigurator.kt
mpawliszyn: Yep updated both of those try/catches. Thanks for the suggestion!

PR #389, opened 2024-05-20
Migrate Authenticated annotations

adrw:
This migrates the annotations to match the upcoming Misk changes.

https://github.com/cashapp/misk/pull/3271

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/CloneBackfillAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/CloneBackfillAction.kt
@@ -90,8 +90,7 @@ class CloneBackfillAction @Inject constructor(
   @Post("/backfills/{id}/clone")
   @RequestContentType(MediaTypes.APPLICATION_JSON)
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
-  // TODO allow any user
-  @Authenticated(capabilities = ["users"])
+  @Authenticated(allowAnyUser = true)
   fun create(
     @PathParam id: Long,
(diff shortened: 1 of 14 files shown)


danieloh0714 (approved):

service/src/main/kotlin/app/cash/backfila/dashboard/CloneBackfillAction.kt
```diff
@@ -90,8 +90,7 @@ class CloneBackfillAction @Inject constructor(
   @Post("/backfills/{id}/clone")
   @RequestContentType(MediaTypes.APPLICATION_JSON)
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
-  // TODO allow any user
-  @Authenticated(capabilities = ["users"])
+  @Authenticated(allowAnyUser = true)
```
mpawliszyn, line 93: This seems wrong? is `@Authenticated` the same as `@Authenticated(capabilities = ["users"])`?

mpawliszyn (approved):

service/src/main/kotlin/app/cash/backfila/dashboard/CloneBackfillAction.kt
adrw: The TODO comment seemed to imply that the desired eventual configuration was any authenticated user but since that didn't exist in Misk until now, allowing the entire `users` group was what was done. That TODO is on every similarly configured annotation in the repo. Open to your / @shellderp 's call on what you want access configured as but this is what I understood from the TODO comment.

shellderp (approved):

PR #390, opened 2024-05-24
Failing dynamo create requests when a range is specified since it is ignored.

mpawliszyn:


client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/internal/DynamoDbBackfillOperator.kt
```diff
@@ -29,6 +29,11 @@ class DynamoDbBackfillOperator<I : Any, P : Any>(
     val config = parametersOperator.constructBackfillConfig(request)
     backfill.validate(config)
 
+    require(
+      request.range == null ||
+        (request.range.start == null && request.range.end == null),
+    ) { "Range is an invalid input for this Dynamo Backfila client" }
```
adrw, line 35:
```suggestion
    ) { "Range is not supported for this Dynamo Backfila client" }
```

client-dynamodb/src/main/kotlin/app/cash/backfila/client/dynamodb/internal/DynamoDbBackfillOperator.kt
```diff
@@ -28,6 +28,11 @@ class DynamoDbBackfillOperator<I : Any, P : Any>(
     val config = parametersOperator.constructBackfillConfig(request)
     backfill.validate(config)
 
+    require(
+      request.range == null ||
+        (request.range.start == null && request.range.end == null),
+    ) { "Range is an invalid input for this Dynamo Backfila client" }
```
adrw, line 34:
```suggestion
    ) { "Range is not supported for this Dynamo Backfila client" }
```

PR #409, opened 2025-01-16
Update `code` element from extending screen width on extra parameters field

Saurinpatel20:
On larger pieces of text, the code element spans the entire width of the screen and extends the text past the width of the page, which makes buttons on the right side of the screen hidden and hard to find without scrolling all the way over. 

Updated the styling on this element so that instead of making it overflow and extend the width of the page, we break the word into separate lines. Used `break-all` as opposed to `break-word`.

### DIFF
--- a/service/web/tabs/app/src/containers/BackfillStatusContainer.tsx
+++ b/service/web/tabs/app/src/containers/BackfillStatusContainer.tsx
@@ -289,7 +289,7 @@ class BackfillStatusContainer extends React.Component<
                       <td>{name}</td>
                       <td>
                       <code style={{ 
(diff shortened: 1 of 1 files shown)


Saurinpatel20:
@mpawliszyn Is it possible to get this merged? It's a quick styling change...

mpawliszyn:
Right, so note that we will be moving to the new UI soon. 
https://github.com/cashapp/backfila/blob/06e2c19a87885ce5916d4af2b06a04d33fc72245/service/src/main/kotlin/app/cash/backfila/ui/components/AlertUseOldUI.kt

Have you run the development app and confirmed this works as you'd expect?

PR #411, opened 2025-01-23
Update Backfill Show to use full width tables

adrw:
Update Backfill Show to use full width tables

Screenshot 2025-01-23 at 16 11 54

Add custom parameters and handling of odd number of rows 

Screenshot 2025-01-27 at 11 52 43


service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -54,6 +55,10 @@ class BackfillShowAction @Inject constructor(
     val label =
       if (backfill.variant == "default") backfill.service_name else "${backfill.service_name} (${backfill.variant})"
 
+    val configurationRows = backfill.toConfigurationRows(id)
```
moshegood, line 58: What if there is an odd number?

mpawliszyn (commented):
Where are the parameter values now?

mpawliszyn (commented):
As mentioned we do need to deal with the odd number of values case. Also where did the parameter information go?

moshegood (approved):

PR #426, opened 2025-03-17
Making the width of the partition a percentage instead of pixels.

mpawliszyn:


iYung (approved):

adrw (approved):
LG

adrw:
Would be great to include a screenshot to confirm this works as expected for the UI change

PR #428, opened 2025-04-03
Update Vitess tests

aparajon:
This PR updates `backfila` to use `VitessTestDb`, Block's newly open-sourced test runner for Vitess which lives in [Misk](https://github.com/cashapp/misk/blob/7751c92cf10c2a2e9a44e9ec13f8e5b9adf4b75a/misk-vitess/src/testFixtures/kotlin/misk/vitess/testing/VitessTestDb.kt#L13). This effectively upgrades the Vitess version from 5.0 to 19.0. Furthermore, the Vitess schema directory structure was also updated to reflect more modern usages, such as omitting the sequence table insert DML's (which are auto-handled by VitessTestDb), removing autoincrements on sharded tables, and fixing vschema linting errors.


adrw (approved):
LG

PR #429, opened 2025-04-07
Show error message on backfill create

adrw:
Screenshot 2025-04-07 at 21 23 49

https://github.com/user-attachments/assets/dc399933-c593-47ab-afe0-148c2f2f76df

This now surfaces any create form failures to the user since this was a point of [frustration](https://cash.slack.com/archives/C0151JAREE7/p1743629121126039).

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
@@ -2,10 +2,14 @@ package app.cash.backfila.ui.actions
 
 import app.cash.backfila.dashboard.CreateBackfillAction
 import app.cash.backfila.protos.service.CreateBackfillRequest
+import app.cash.backfila.ui.components.AlertError
+import app.cash.backfila.ui.components.DashboardPageLayout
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
```diff
@@ -55,11 +60,23 @@ class BackfillCreateHandlerAction @Inject constructor(
...
-      variant = formFields[BackfillCreateField.VARIANT.fieldId]!!,
-      request = createRequestBuilder.build(),
-    )
+    val response = try {
+      createBackfillAction.create(
+        service = formFields[BackfillCreateField.SERVICE.fieldId]!!,
+        variant = formFields[BackfillCreateField.VARIANT.fieldId]!!,
+        request = createRequestBuilder.build(),
+      )
+    } catch (e: BadRequestException) {
```
mpawliszyn, line 69: Wouldn't this be a cash most/all errors situation?

adrw: Yes, and then it would show them in the UI. Since this action only handles form submissions from the UI, I think that is the desired functionality.

PR #430, opened 2025-04-09
Show error in UI when update field fails

adrw:
Bump Misk 2025.04.09.202256-90dc722 which fixes old Backfila UI failing
on missing web assets


service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillShowButtonHandlerAction.kt
```diff
@@ -35,43 +39,63 @@ class BackfillShowButtonHandlerAction @Inject constructor(
...
+      // Since this action is only hit from the UI, catch any validation errors and show them to the user
+      val errorHtmlResponseBody = dashboardPageLayout.newBuilder()
+        .buildHtmlResponseBody {
+          div("py-20") {
+            AlertError(message = "Update backfill field failed: $e", label = "Try Again", onClick = "history.back(); return false;")
+          }
+        }
+      return Response(
+        body = errorHtmlResponseBody,
+        statusCode = 200,
```
sq-js, line 96: statusCode 200 for an exception? 🤔

adrw: Yes because this endpoint is only serving browser traffic

sq-js (approved):

PR #432, opened 2025-04-11
Add three range logic to new form - Restart, continue and new

jeffhwang-sq:
The logic for clone (create stays the same) - 
CONTINUE - use cursor if available, failing back to start position 
RESTART - use original start position
NEW - use form values (this is what we only had) 

https://github.com/user-attachments/assets/eb23384f-2f09-4794-ac54-383d8710b5c1

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
@@ -131,9 +131,5 @@ class BackfillCreateHandlerAction @Inject constructor(
     const val PATH = "/api/backfill/create"
 
     private fun <T>String?.ifNotBlank(block: (String) -> T) = if (this.isNullOrBlank()) null else block(this)
-
-    private fun getBackfillIdFromName(name: String): Long? {
-      return name.toLongOrNull()
-    }
   }
 }
--- a/service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
@@ -56,7 +56,7 @@ class BackfillCreateHandlerAction @Inject constructor(
       when (formFields[BackfillCreateField.RANGE_OPTION.fieldId]) {
         RangeOption.CONTINUE.value -> {
           // Get the last processed position from the original backfill
-          val backfillId = formFields[BackfillCreateField.CLONE_BACKFILL_ID.fieldId]?.toLongOrNull()
+          val backfillId = formFields[BackfillCreateField.BACKFILL_ID_TO_CLONE.fieldId]?.toLongOrNull()
           backfillId?.let { id ->
             val status = getBackfillStatusAction.status(id)
             status.partitions.firstOrNull()?.let { partition ->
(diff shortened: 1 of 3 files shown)


jeffhwang-sq:
I am trying to build s3 test service - wip

service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
```diff
@@ -48,8 +51,39 @@ class BackfillCreateHandlerAction @Inject constructor(
...
         },
       )
-      formFields[BackfillCreateField.RANGE_START.fieldId]?.ifNotBlank { createRequestBuilder.pkey_range_start(it.encodeUtf8()) }
-      formFields[BackfillCreateField.RANGE_END.fieldId]?.ifNotBlank { createRequestBuilder.pkey_range_end(it.encodeUtf8()) }
+
+      // Handle range options for cloned backfills
+      when (formFields[BackfillCreateField.RANGE_OPTION.fieldId]) {
+        RangeOption.CONTINUE.value -> {
+          // Get the last processed position from the original backfill
+          val backfillId = formFields[BackfillCreateField.BACKFILL_NAME.fieldId]?.toLongOrNull()
```
adrw, line 59: This is going to be the fully qualified name of the backfill class, not the DB ID as a Long so this parse will always resolve to null. You'll need to add BACKFILL_ID_TO_CLONE as a new enum and hidden input field to fill in the create action so you can get it in the create handler.

service/src/main/kotlin/app/cash/backfila/ui/actions/BackfillCreateHandlerAction.kt
```diff
@@ -97,5 +131,9 @@ class BackfillCreateHandlerAction @Inject constructor(
     const val PATH = "/api/backfill/create"
 
     private fun <T>String?.ifNotBlank(block: (String) -> T) = if (this.isNullOrBlank()) null else block(this)
+
+    private fun getBackfillIdFromName(name: String): Long? {
+      return name.toLongOrNull()
```
adrw, line 136: This doesn't make any sense, lmk when you want a re-review

mpawliszyn: I don't think this is even used?

jeffhwang-sq: Right, this is not used. Bringing the id directly now. Removed this function.

service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillCreateAction.kt
```diff
@@ -194,38 +200,128 @@ class BackfillCreateAction @Inject constructor(
...
+                    // Range fields for new backfills (non-clone case)
+                    div("mt-10 grid grid-cols-1 gap-x-6 gap-y-8 sm:grid-cols-6") {
+                      div("sm:col-span-3") {
+                        val field = BackfillCreateField.RANGE_START.fieldId
+                        label("block text-sm/6 font-medium text-gray-900") {
+                          htmlFor = field
+                          +"""Range Start (optional)"""
+                        }
+                        div("mt-2") {
+                          input(classes = "block w-full rounded-md bg-white px-3 py-1.5 text-base text-gray-900 outline outline-1 -outline-offset-1 outline-gray-300 placeholder:text-gray-400 focus:outline focus:outline-2 focus:-outline-offset-2 focus:outline-indigo-600 sm:text-sm/6") {
```
mpawliszyn, line 304: @adrw Are there good ways to handle all this copy and paste? I wonder if we should be doing something about this or not?

mpawliszyn (approved):
Overall though, looks fine.

PR #433, opened 2025-04-14
Log create/clone/update errors to Datadog, not just browser UI

jeffhwang-sq:
Previously, error would only show up in browser UI, but no stacktrace in Datadog / logs provider to be able to investigate fully.


adrw (approved):
LG

PR #434, opened 2025-04-14
Make menu links configurable

adrw:
Now internal deployments can add their own custom links which will be added to the menu.

Screenshot 2025-04-14 at 15 59 35


jeffhwang-sq:
Looks good. Thanks for pairing as well.

PR #440, opened 2025-04-22
Enable ETA on the new UI 

jeffhwang-sq:
Screenshot 2025-04-21 at 8 44 12 PM
Screenshot 2025-04-21 at 8 44 36 PM
Screenshot 2025-04-21 at 8 44 45 PM

Empty State Messages:
- "Computing size..." when precomputing isn't done
- "-" when not in RUNNING state
- "Calculating..." when we don't have enough data to calculate ETA

ETA Calculation
- Use the time formatting from the old UI (https://github.com/cashapp/backfila/blob/6591e0dc8b9736372aff8a324dde6e1cfc30bb9d/service/web/tabs/app/src/containers/BackfillStatusContainer.tsx#L55)

The ETA will show:
- "Computing size..." during precomputation
- "-" when not running
- "Calculating..." when running but missing required data
-  "1h30m" when we have all required data


service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -178,8 +178,18 @@ class BackfillShowAction @Inject constructor(
                       )
                     }
                     td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") { +"""${partition.matching_records_per_minute} #/m""" }
-                    // TODO properly calculate the ETA until finished
-                    td("py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700") { +"""ETA TODO""" }
+                    td("py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700") {
+                      when {
+                        partition.state != BackfillState.RUNNING -> +"-"
+                        !partition.precomputing_done -> +"Computing size..."
```
adrw, line 184:
```suggestion
                        !partition.precomputing_done -> +"Computing..."
```

Less text is better given it is a narrow column

adrw (approved):
This is awesome, nice copy over of formatting logic

PR #441, opened 2025-04-22
Add proper state handling for rate display on the new UI

jeffhwang-sq:
When precomputing
Screenshot 2025-04-21 at 9 14 25 PM

When precomputing is done, but no rate to provide
Screenshot 2025-04-21 at 8 55 17 PM

When in running state
Screenshot 2025-04-21 at 8 55 37 PM

When complete
Screenshot 2025-04-21 at 8 55 47 PM

Empty State Messages:
- "Computing size..." when precomputing isn't done
- "-" when not in RUNNING state
- "Calculating..." when we don't have enough data to calculate ETA

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
@@ -180,7 +180,7 @@ class BackfillShowAction @Inject constructor(
                     td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") {
                       when {
                         partition.state != BackfillState.RUNNING -> +"-"
-                        !partition.precomputing_done -> +"Computing size..."
+                        !partition.precomputing_done -> +"Computing..."
                         partition.matching_records_per_minute == null ||
                           partition.matching_records_per_minute <= 0 -> +"Calculating..."
                         else -> +"""${partition.matching_records_per_minute} #/m"""
--- a/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
@@ -177,7 +177,15 @@ class BackfillShowAction @Inject constructor(
                         partition.computed_matching_record_count,
                       )
                     }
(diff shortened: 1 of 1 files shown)


service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -177,7 +177,15 @@ class BackfillShowAction @Inject constructor(
                         partition.computed_matching_record_count,
                       )
                     }
-                    td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") { +"""${partition.matching_records_per_minute} #/m""" }
+                    td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") {
+                      when {
+                        partition.state != BackfillState.RUNNING -> +"-"
+                        !partition.precomputing_done -> +"Computing size..."
```
adrw, line 183:
```suggestion
                        !partition.precomputing_done -> +"Computing..."
```

Less text is better for narrow column

adrw (approved):
LG

service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -177,7 +177,15 @@ class BackfillShowAction @Inject constructor(
                         partition.computed_matching_record_count,
                       )
                     }
-                    td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") { +"""${partition.matching_records_per_minute} #/m""" }
+                    td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") {
+                      when {
+                        partition.state != BackfillState.RUNNING -> +"-"
+                        !partition.precomputing_done -> +"Computing..."
```
mpawliszyn, line 183: Records per minute shouldn't depend on the precomputing step. That is literally the rate of runBatch only and those should start right away.

jeffhwang-sq: when I tested this out, the values were always 0 when precomputing was set to False, then when its set to True, the value gets filled.
Screenshot 2025-04-22 at 10 57 19 AM
Screenshot 2025-04-22 at 10 56 38 AM

PR #443, opened 2025-04-28
Enable Soft Deletion Backfills

jeffhwang-sq:
Allow soft-delete backfills for cancelled and completed backfills. 

Soft Deletion Handling: Backfila is updated to recognize and process records flagged as soft-deleted, ensuring they are included or excluded from backfills based on configuration.​

A 'Delete' button has been added to cancelled or completed backfills in the UI, providing users with the ability to soft delete backfills.

We filter the deleted backfills from the UI.

https://github.com/user-attachments/assets/2f755a1f-b0c7-4b95-a515-8b6d000dc320

Modified to nullable timestamp
Screenshot 2025-04-29 at 1 01 39 PM

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
@@ -26,7 +26,7 @@ class SoftDeleteBackfillAction @Inject constructor(
   @Post("/backfill/delete/{id}")
   @RequestContentType(MediaTypes.APPLICATION_JSON)
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
-  @Authenticated
+  @Authenticated(capabilities = ["users"])
   fun softDelete(
     @PathParam id: Long,
   ) {
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
@@ -38,7 +38,7 @@ class SoftDeleteBackfillAction @Inject constructor(
         throw BadRequestException("Can only delete completed or cancelled backfills")
       }
 
-      backfillRun.soft_deleted = true
+      backfillRun.deleted_at = java.time.Instant.now()
 
       // Log the deletion event
       session.save(
--- a/service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
@@ -1,0 +1,55 @@
+package app.cash.backfila.dashboard
+
+import app.cash.backfila.service.persistence.BackfilaDb
+import app.cash.backfila.service.persistence.BackfillState
+import app.cash.backfila.service.persistence.DbBackfillRun
+import app.cash.backfila.service.persistence.DbEventLog
+import javax.inject.Inject
+import misk.MiskCaller
+import misk.exceptions.BadRequestException
+import misk.hibernate.Id
+import misk.hibernate.Transacter
+import misk.hibernate.load
+import misk.scope.ActionScoped
+import misk.security.authz.Authenticated
+import misk.web.PathParam
+import misk.web.Post
+import misk.web.RequestContentType
+import misk.web.ResponseContentType
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+
+class SoftDeleteBackfillAction @Inject constructor(
+  @BackfilaDb private val transacter: Transacter,
+  private val caller: @JvmSuppressWildcards ActionScoped<MiskCaller?>,
+) : WebAction {
+  @Post("/backfill/delete/{id}")
+  @RequestContentType(MediaTypes.APPLICATION_JSON)
+  @ResponseContentType(MediaTypes.APPLICATION_JSON)
+  @Authenticated
+  fun softDelete(
+    @PathParam id: Long,
+  ) {
+    transacter.transaction { session ->
+      val backfillRun = session.load<DbBackfillRun>(Id(id))
+
+      // Only allow soft delete for COMPLETE or CANCELLED backfills
+      if (backfillRun.state != BackfillState.COMPLETE && backfillRun.state != BackfillState.CANCELLED) {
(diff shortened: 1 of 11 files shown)


service/src/main/resources/migrations/v020__add_soft_delete.sql
```diff
@@ -0,0 +1,3 @@
+-- Add soft_deleted column to backfill_runs
+ALTER TABLE backfill_runs
+    ADD COLUMN soft_deleted BOOLEAN NOT NULL DEFAULT FALSE;
```
adrw, line 3: Will this fill in false for all existing rows? Will queries fail for existing backfills which have null instead of false?

service/src/main/kotlin/app/cash/backfila/service/persistence/DbBackfillRun.kt
```diff
@@ -90,6 +90,9 @@ class DbBackfillRun() : DbUnsharded<DbBackfillRun>, DbTimestampedEntity {
   @Column(nullable = false)
   var dry_run: Boolean = false
 
+  @Column(nullable = false)
+  var soft_deleted: Boolean = false
```
adrw, line 94: I'd lean towards a `deleted_at` with a timestamp which is by default null, it would make future pruning easier than a simple boolean if that's an anticipated maintenance consideration.

service/src/main/kotlin/app/cash/backfila/service/persistence/BackfillSoftDelete.kt
```diff
@@ -0,0 +1,5 @@
+package app.cash.backfila.service.persistence
+
+enum class BackfillSoftDelete {
```
adrw, line 3: What would the other states be? If no other planned states, why is this Enum necessary?

adrw (approved):
LG

service/src/main/kotlin/app/cash/backfila/dashboard/SoftDeleteBackfillAction.kt
```diff
@@ -0,0 +1,55 @@
...
+import misk.web.mediatype.MediaTypes
+
+class SoftDeleteBackfillAction @Inject constructor(
+  @BackfilaDb private val transacter: Transacter,
+  private val caller: @JvmSuppressWildcards ActionScoped<MiskCaller?>,
+) : WebAction {
+  @Post("/backfill/delete/{id}")
+  @RequestContentType(MediaTypes.APPLICATION_JSON)
+  @ResponseContentType(MediaTypes.APPLICATION_JSON)
+  @Authenticated
```
adrw, line 29:
```suggestion
  @Authenticated(capabilities = ["users"])
```

Empty authenticated will fail in real environments, always set capabilities or services.

service/src/main/resources/migrations/v020__add_soft_delete.sql
jeffhwang-sq: Yeah, this will modify all existing rows by adding the soft_deleted column and setting its value to FALSE for each of them. So the queries will not fail. Also when inserting a new row without specifying the soft_deleted column, the database will automatically use the default value, which is FALSE.

service/src/main/kotlin/app/cash/backfila/service/persistence/BackfillSoftDelete.kt
jeffhwang-sq: This is the only state actually. Yeah, maybe enum is overdoing here. Will just use a simple string.

service/src/main/kotlin/app/cash/backfila/service/persistence/DbBackfillRun.kt
jeffhwang-sq: I did not consider that since boolean has simpler logic, lightweight storage and fast indexing. But if we consider pruning and having some hard deletion logic, having timestamp would be helpful.

service/src/main/kotlin/app/cash/backfila/service/persistence/BackfillSoftDelete.kt
jeffhwang-sq: I started making the change, but since this field is used in multiple places, it might be better to define it as an enum. That way, the states are centralized and easier to manage consistently across the codebase.

service/src/main/kotlin/app/cash/backfila/service/persistence/DbBackfillRun.kt
jeffhwang-sq: Modified to nullable timestamp

service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -287,13 +290,27 @@ class BackfillShowAction @Inject constructor(
...
   }
 
+  private fun getDeleteButton(state: BackfillState, deletedAt: Instant?): Link? {
+    if (deletedAt != null) {
+      return null
+    }
+    return when (state) {
+      BackfillState.COMPLETE, BackfillState.CANCELLED -> Link(
+        label = DELETE_STATE_BUTTON_LABEL,
+        href = BackfillSoftDelete.SOFT_DELETED.name,
```
mpawliszyn, line 300: Just show the deleted at rather than expose this enum that is a string.

mpawliszyn: "Soft Delete"

service/src/main/resources/migrations/v020__add_soft_delete.sql
```diff
@@ -0,0 +1,3 @@
+-- Add soft_deleted column to backfill_runs
+ALTER TABLE backfill_runs
```
mpawliszyn, line 2: Delete this.

service/src/main/resources/migrations/v019__backfila.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE backfill_runs
+    ADD COLUMN `deleted_at` TIMESTAMP NULL DEFAULT NULL;
```
mpawliszyn, line 2: Follow the existing formatting
`  deleted_in_service_at timestamp(3) NULL DEFAULT NULL,`

mpawliszyn (approved):

mpawliszyn:
Make sure to follow up with updating the backfill page with the deleted at so there is some feedback when hitting delete.

PR #444, opened 2025-04-28
enable toggle view for soft deleted backfill runs

jeffhwang-sq:
New soft_delete_column option:
- Backfill clients can now specify a soft_delete_column in the @BackfillConfig.
- Rows where the soft_delete_column is null are treated as active and eligible for backfill.
- With toggle, users can hide soft deleted backfills in the dashboard.

https://github.com/user-attachments/assets/dcb4e7ef-8fbd-4785-a2d6-4a66c8965ef2

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/pages/ServiceShowAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/pages/ServiceShowAction.kt
@@ -83,7 +83,7 @@ class ServiceShowAction @Inject constructor(
             }
           }
 
-          BackfillsTable(true, backfillRuns.running_backfills, showDeleted)
+          BackfillsTable(true, backfillRuns.running_backfills)
           BackfillsTable(false, backfillRuns.paused_backfills, showDeleted)
           Pagination(backfillRuns.next_pagination_token, offset, lastOffset, path(service, variantOrBlank))
         }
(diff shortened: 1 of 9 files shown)


service/src/main/kotlin/app/cash/backfila/ui/pages/ServiceShowAction.kt
```diff
@@ -70,16 +76,15 @@ class ServiceShowAction @Inject constructor(
...
               button(classes = "rounded-full bg-indigo-600 px-3 py-1.5 text-sm font-semibold text-white shadow-sm hover:bg-indigo-500 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-indigo-600") {
                 type = ButtonType.button
                 +"""Create"""
               }
             }
           }
 
-          BackfillsTable(true, backfillRuns.running_backfills)
-          BackfillsTable(false, backfillRuns.paused_backfills)
+          BackfillsTable(true, backfillRuns.running_backfills, showDeleted)
```
adrw, line 86: I thought you only wanted to show the toggle for paused backfills? But it controls the running ones as well?

adrw (approved):
LG

service/src/main/kotlin/app/cash/backfila/ui/pages/ServiceShowAction.kt
jeffhwang-sq: we do ignore the toggle for running backfills in backfills table let let me clarify by removing the showDeleted as the third parameter

mpawliszyn (approved):

PR #447, opened 2025-05-02
Fix DynamoDB BatchWriteItem 25-item limit in UpdateInPlaceDynamoDbBackfill

notmikedavis:
# Problem
While running a backfill, encountered a `DynamoDbException` with the error:
```
[Member must have length less than or equal to 25, Member must have length greater than or equal to 1]
```
This occurs because the AWS DynamoDB BatchWriteItem API has a strict limit of 25 items per request, but the current implementation of `UpdateInPlaceDynamoDbBackfill` doesn't account for this limit when building the batch request. When a backfill attempts to process more than 25 items in a single batch, it fails with the above exception.

# Solution
Modified `UpdateInPlaceDynamoDbBackfill` to process items in chunks of 25 or less using Kotlin's `chunked()` function. Each chunk is processed with its own `BatchWriteItemRequest`, ensuring we never exceed the AWS DynamoDB BatchWriteItem limit.

Key changes:
- Added a `BATCH_SIZE_LIMIT` constant set to 25
- Use `chunked(BATCH_SIZE_LIMIT)` to split items into appropriate batch sizes
- Process each chunk with a separate `BatchWriteItemRequest`


jeffhwang-sq:
Thanks for doing this. Change looks good to me. Surprised this was never raised before.

jeffhwang-sq (approved):

PR #448, opened 2025-05-05
Jhwang/add ability to modify partition

jeffhwang-sq:
Enabe dynamic partition adjustments for paused backfills 
- Adds a new option to edit pkey cursor.
- Adds a new landing page to adjust the pkey. 
- Contain warning and success message. 

https://github.com/user-attachments/assets/32ea330d-184c-4cee-b5d5-0aa1617a1d3e

Screenshot 2025-05-05 at 12 56 57 PM

Screenshot 2025-05-05 at 12 57 29 PM

Screenshot 2025-05-08 at 12 19 40 PM


adrw:
On success, it should redirect back to the Backfill Show page.

jeffhwang-sq:
> On success, it should redirect back to the Backfill Show page. Alternatively it'd be nice to have the change form inline like the other config change buttons which turn into input boxes.

Changes are done but let me see if I can add some tests.

service/src/main/kotlin/app/cash/backfila/dashboard/EditPartitionCursorHandlerAction.kt
```diff
@@ -0,0 +1,91 @@
...
+  ): Response<ResponseBody> {
+    val request = httpCall.get().asOkHttpRequest()
+    val cursorSnapshot = request.url.queryParameter("cursor_snapshot")?.takeIf { it.isNotBlank() }
+    val newCursor = request.url.queryParameter("new_cursor")
+
+    // Validate UTF-8
+    try {
+      newCursor?.toByteArray(Charsets.UTF_8)?.toString(Charsets.UTF_8)
+    } catch (e: Exception) {
+      throw BadRequestException("New cursor must be valid UTF-8")
```
adrw, line 50: All of these BadRequestExceptions are going to fail without showing UI to the frontend. Can you update to have these errors surface in a error banner?

Something like this: https://github.com/cashapp/backfila/pull/430/files

service/src/test/kotlin/app/cash/backfila/actions/EditPartitionEndPointTest.kt
```diff
@@ -0,0 +1,162 @@
...
+
+  @Inject lateinit var queryFactory: Query.Factory
+
+  @Inject lateinit var scope: ActionScope
+
+  /**
+   * Test-specific action that provides the same functionality as EditPartitionCursorHandlerAction
+   * but without the HTTP dependencies
+   */
+  inner class TestEditPartitionAction {
```
adrw, line 55: So the test is only testing this fake action? That doesn't give us much. We should refactor the actual EditParitionCursorAction so it's testable, or test it as is handling the HTTP dependencies, or delete this test.

adrw (approved):

service/src/main/kotlin/app/cash/backfila/dashboard/EditPartitionCursorHandlerAction.kt
jeffhwang-sq: Sounds good. I will keep this in mind. Keep thinking by throwing, users will be able to see the error message.

jeffhwang-sq: updated the description with the actual error to the UI

PR #449, opened 2025-05-05
Improve BatchWriteItem retry handling

notmikedavis:
## Problem

The DynamoDB BatchWriteItem API often has transient failures with unprocessed items that cause entire Backfila batches to fail. These could be more granularly retried within a run to avoid Backfila getting stuck on a batch.

## Solution

Improve the retry mechanism to better align with [AWS BatchWriteItem best practices](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html):
- Collect unprocessed items across all batches and retry them together
- Use exponential backoff with jitter to handle throttling
- Only count towards retry limit when no progress is made
- Provide more detailed error reporting

### DIFF
--- a/client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
+++ b/client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
@@ -3,55 +3,115 @@ package app.cash.backfila.client.dynamodbv2
 import app.cash.backfila.client.BackfillConfig
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemRequest
+import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemResponse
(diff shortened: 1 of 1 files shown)


client-dynamodb-v2/src/main/kotlin/app/cash/backfila/client/dynamodbv2/UpdateInPlaceDynamoDbBackfill.kt
```diff
@@ -3,60 +3,122 @@ package app.cash.backfila.client.dynamodbv2
...
 abstract class UpdateInPlaceDynamoDbBackfill<I : Any, P : Any>(
   val dynamoDbClient: DynamoDbClient,
-
 ) : DynamoDbBackfill<I, P>() {
   companion object {
     // DynamoDB BatchWriteItem API has a limit of 25 items per request
     private const val BATCH_SIZE_LIMIT = 25
+    
+    // Retry configuration
+    private const val MAX_RETRY_ATTEMPTS = 10
```
jeffhwang-sq, line 28: I am not sure if all the failures are transient failures. If not, retry of 10 may be too much. I will try to add some loggings after this pr merges so we can see how many failures we encounter.

jeffhwang-sq (approved):

PR #452, opened 2025-05-06
Add ApiCallTimeoutException handling to UpdateInPlaceDynamoDbBackfill

notmikedavis:
## Problem

The DynamoDB BatchWriteItem implementation in UpdateInPlaceDynamoDbBackfill currently lacks handling for ApiCallTimeoutException. When these timeouts occur, the entire batch fails without any retry attempts, causing backfills to fail unnecessarily.

## Solution

Add comprehensive timeout handling with these features:
- Track and retry chunks that experience API timeouts
- Use exponential backoff with jitter for retries
- Only increment the timeout counter when all chunks in an iteration timeout
- Reset the timeout counter if any chunk succeeds
- Maintain separate counters for timeouts vs unprocessed items
- Provide detailed error context through suppressed exceptions

The implementation is designed to be resilient to transient timeouts while still protecting against systemic failures. It coordinates the backoff strategy between timeout retries and unprocessed item retries.


jeffhwang-sq (approved):

jeffhwang-sq:
Approving to unblock, need to add some loggings.

PR #453, opened 2025-05-07
enable non numerical values for range

jeffhwang-sq:
The changes involve modifications to the range-handling logic, allowing Backfila to process start and end range to be non-numeric values.

This change addresses a discussion raised in [Slack](https://cash.slack.com/archives/C0151JAREE7/p1746490372275049). The backfill at [Backfila #17385](https://backfila.stage.sqprod.co/backfills/17385) uses Jooq backfills. Depending on the specific type of backfill, we want to support non-numerical range values and delegate range validation to the corresponding backfill operator. For instance, the StaticDatasourceBackfillOperator will enforce that ranges are numerical.

Will save this for Mike to review


mpawliszyn (approved):
Yep, this is right. There is an assumption here around being utf8 encoded but this retains the existing functionality of the old UI.

PR #458, opened 2025-05-14
Add last registered at column to Services

jeffhwang-sq:
We're adding a new last_registered_at column to the services table. This is similar to the updated_at column in registered_backfills, but instead of tracking only config changes, it will also update whenever the service is re-registered.
This column will be used in our deprecation reminder. 

When reregistered
Screenshot 2025-05-14 at 3 27 40 PM

When registered
Screenshot 2025-05-14 at 3 27 17 PM

### DIFF
--- a/service/src/main/resources/migrations/v020__backfila.sql
+++ b/service/src/main/resources/migrations/v020__backfila.sql
@@ -1,0 +1,2 @@
+ALTER TABLE services
+ADD COLUMN last_registered_at timestamp NULL;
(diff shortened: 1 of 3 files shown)


service/src/main/resources/migrations/v020__backfila.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE services
+ADD COLUMN last_registered_at timestamp NULL;
```
mpawliszyn, line 2: We could set a default:
DEFAULT NOW(3)

PR #463, opened 2025-05-21
Adjust rate logic to render even in precomputing stage

jeffhwang-sq:
Fix a discrepancy in how backfill rates are displayed between the old and new UI implementations.

## Issue
The new UI was incorrectly hiding valid rate information during precomputing phase. For example:
- Old UI: Shows "300 #/m" (correct)
- New UI: Shows "Computing..." (incorrect when we have a valid rate)
- When the pre-computation is done, the rate matches with the old UI
## Changes
Modified the rate display logic in BackfillShowAction.kt to:
1. Show actual rates when available, even during precomputing
2. Only show "Computing..." when there's no rate during precomputing
3. Show "Calculating..." when there's no rate after precomputing is done

https://backfila.sqprod.co/backfills/29757
Old UI renders rate 
Screenshot 2025-05-21 at 1 22 02 PM

New UI Does not
Screenshot 2025-05-21 at 1 21 58 PM

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
@@ -194,8 +194,7 @@ class BackfillShowAction @Inject constructor(
                         partition.state != BackfillState.RUNNING -> +"-"
                         partition.matching_records_per_minute == null ||
                           partition.matching_records_per_minute <= 0 -> {
-                            if (!partition.precomputing_done) +"Computing..."
-                            else +"Calculating..."
+                          if (!partition.precomputing_done) { +"Computing..." } else { +"Calculating..." }
                         }
                         else -> +"""${partition.matching_records_per_minute} #/m"""
                       }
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -192,9 +192,11 @@ class BackfillShowAction @Inject constructor(
                     td("hidden py-5 pl-8 pr-0 text-right align-top tabular-nums text-gray-700 sm:table-cell") {
                       when {
                         partition.state != BackfillState.RUNNING -> +"-"
-                        !partition.precomputing_done -> +"Computing..."
                         partition.matching_records_per_minute == null ||
-                          partition.matching_records_per_minute <= 0 -> +"Calculating..."
+                          partition.matching_records_per_minute <= 0 -> {
+                            if (!partition.precomputing_done) +"Computing..."
```
adrw, line 197:
```suggestion
                            if (!partition.precomputing_done) { +"Computing..." }
```

service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -192,9 +192,11 @@ class BackfillShowAction @Inject constructor(
...
+                            else +"Calculating..."
```
adrw, line 198:
```suggestion
                            else { +"Calculating..." }
```

adrw (approved):
Braces for multi-line if statements pls

PR #464, opened 2025-05-21
Fix progress bar to match the old ui

jeffhwang-sq:
# Fix misleading progress bar behavior in new UI

## Issue
The new UI's progress bar implementation has a bug where it shows misleading progress during precomputing:
- Shows 100% initially when denominator is 0
- Then jumps to actual progress (e.g., 1%) as computation proceeds
- Doesn't match old UI's behavior of showing indeterminate state during precomputing

## Changes
Updated ProgressBar component to match old UI's behavior:

https://github.com/user-attachments/assets/135ce16b-6946-4244-b0d9-dfe1a2ed9663

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/components/ProgressBar.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/components/ProgressBar.kt
@@ -21,7 +21,7 @@ fun TagConsumer<*>.ProgressBar(
     if (percentComplete == null) {
       div("animate-pulse bg-gray-300 text-xs font-medium text-center p-0.5 leading-none rounded-full") {
         style = "width: 100%"
-        +"Computing..."
+        +"..."
       }
     } else {
       // Don't show blue sliver for 0%, just show the gray empty bar
--- a/service/src/main/kotlin/app/cash/backfila/ui/components/ProgressBar.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/components/ProgressBar.kt
@@ -5,16 +5,31 @@ import kotlinx.html.TagConsumer
 import kotlinx.html.div
 import kotlinx.html.style
 
-fun TagConsumer<*>.ProgressBar(numerator: Number, denominator: Number) {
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/app/cash/backfila/ui/components/ProgressBar.kt
```diff
@@ -5,16 +5,31 @@ import kotlinx.html.TagConsumer
...
+      !precomputingDone -> null // Show indeterminate during precomputing
+      denominator.toDouble() <= 0 -> 0.0 // Handle zero denominator
+      else -> round((numerator.toDouble() / denominator.toDouble()) * 100)
+    }
+
+    // Show indeterminate progress bar during precomputing
+    if (percentComplete == null) {
+      div("animate-pulse bg-gray-300 text-xs font-medium text-center p-0.5 leading-none rounded-full") {
+        style = "width: 100%"
+        +"Computing..."
```
adrw, line 24:
```suggestion
        +"..."
```

Not sure we need "computing" in 3 spots in the row since it shows up in the cells beside it too

service/src/main/kotlin/app/cash/backfila/ui/components/ProgressBarReloader.kt
```diff
@@ -5,16 +5,31 @@ import kotlinx.html.TagConsumer
...
+      !precomputingDone -> null // Show indeterminate during precomputing
+      denominator.toDouble() <= 0 -> 0.0 // Handle zero denominator
+      else -> round((numerator.toDouble() / denominator.toDouble()) * 100)
+    }
+
+    // Show indeterminate progress bar during precomputing
+    if (percentComplete == null) {
+      div("animate-pulse bg-gray-300 text-xs font-medium text-center p-0.5 leading-none rounded-full") {
+        style = "width: 100%"
+        +"Computing..."
```
adrw, line 24:
```suggestion
        +"..."
```

adrw (approved):
LG

PR #469, opened 2025-06-09
Add turbo frame to events and partition

jeffhwang-sq:
Enable turbo-frame for auto-reload

Use turbo-frame to implement partial page updates for backfill status. This allows the partitions and events sections to auto-reload while keeping the configuration section stable.

Changes:
- Add turbo-frame support in AutoReload component
- Configure frame attributes for proper reloading
- Update auto-reload controller to use Turbo.visit for frame updates

https://github.com/user-attachments/assets/b22fff0b-c12b-432d-9a3d-61c90a5f79cc

### DIFF
--- a/service/src/main/resources/web/static/js/auto_reload_controller.js
+++ b/service/src/main/resources/web/static/js/auto_reload_controller.js
@@ -1,41 +1,27 @@
 import { Application, Controller } from "../cache/stimulus/3.1.0/stimulus.min.js";
 window.Stimulus = Application.start();
 
-// Hard reload
 Stimulus.register("auto-reload", class extends Controller {
+  static targets = ["frame"]
+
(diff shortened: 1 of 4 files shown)


service/src/main/resources/web/static/js/auto_reload_controller.js
```diff
@@ -1,41 +1,27 @@
...
-//  }
-//
-//  reloadFrame() {
-//    this.frameTarget.src = this.frameTarget.src;
-//  }
-//});
\ No newline at end of file
+  startReloading() {
+    this.interval = setInterval(() => {
+      console.log("auto-reload triggered...");
```
adrw, line 19:
```suggestion
```

Don't log in production if we're not debugging

PR #470, opened 2025-06-10
Move state section to its own Turbo Frame

jeffhwang-sq:
# Move state section to its own Turbo Frame

## Changes

- Moved state section to its own Turbo Frame with dedicated auto-reload
- Extracted button rendering logic into reusable functions in `BackfillShowButtonHandlerAction`
- Removed duplicate button helper functions from `BackfillShowAction`
- Added proper frame response handling for state updates

https://github.com/user-attachments/assets/4eb462ea-9cb1-470b-860b-003680c96b9d


service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -102,6 +103,24 @@ class BackfillShowAction @Inject constructor(
           }
         }
 
+        // State section with its own auto-reload
+        AutoReload(frameId = "backfill-$id-state") {
+          Card {
+            turbo_frame("backfill-$id-state") {
```
keeferrourke, line 109: AutoReload already includes a turbo_frame with this ID. I think this should be removed?

PR #495, opened 2025-07-18
Modify turbo frame issue with simple fetch based approach

jeffhwang-sq:
Pagination Auto-Reload Issue:
  -  Replaced complex Turbo-based reload with simple fetch-based approach
  - Cleanly strips frame parameters from URLs to prevent contamination
  - Preserves pagination and filter parameters during auto-reload
  - Works consistently for all frame types (status frames, state buttons, etc.)

https://github.com/user-attachments/assets/1c333cce-3a64-41b4-800c-adea721261f4

### DIFF
--- a/service/src/main/resources/web/static/js/auto_reload_controller.js
+++ b/service/src/main/resources/web/static/js/auto_reload_controller.js
@@ -5,22 +5,42 @@ Stimulus.register("auto-reload", class extends Controller {
   static targets = ["frame"]
 
   connect() {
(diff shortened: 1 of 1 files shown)


adrw (approved):
LG, let's give this a shot but revert if we end up seeing other issues with this approach.

jeffhwang-sq:
> LG, let's give this a shot but revert if we end up seeing other issues with this approach.

Sounds good! Might need to add more logic for stack trace change.

PR #498, opened 2025-07-22
Keep stack trace expansion state with auto upload

jeffhwang-sq:
When users expanded event logs using "Show More" to view full stack traces or error details, the auto-reload functionality would reset all expanded states back to collapsed every 10 seconds.

This pr enables state preservation for event log expansions during Turbo frame auto-reloads.

https://github.com/user-attachments/assets/11eb02c9-5da5-4cdb-8590-a7966d1c030e

Fixing the snapping issue

https://github.com/user-attachments/assets/dddd67ec-0053-4c4d-8099-e470a7d939fb


jeffhwang-sq:
> Why is is snapping left when you open details?

Oh I think thats happening when I click to the next page?

jeffhwang-sq:
> Why is is snapping left when you open details?

Looks like I need to set explicit width property.

adrw (approved):
LG

PR #500, opened 2025-07-24
Only keep max width for shards and logs

jeffhwang-sq:
Configuration and Overall Progress sections: Use the original constrained width
Partitions and Events tables: Use full width to accommodate data tables that benefit from more horizontal space (These are centered now)

Regular backfills with shorter logs and events - you can see that the contents are centered same with config and overall progress. The cards are full width still to accommodate the other error cases.

https://github.com/user-attachments/assets/02d75dfe-bf69-4e0e-ae0d-4a6b2bffc14e

Backfill with longer error messages. The content are still centered but utilizes the empty spaces now. 
https://github.com/user-attachments/assets/9b4d3a5f-d76a-404a-a829-5c38b36d4e6c

Small laptop screen still enables scrolling 

https://github.com/user-attachments/assets/191e3fd6-8c71-4b1c-bd2c-b72150d54bf0

Small laptop screen without errors

https://github.com/user-attachments/assets/737a7849-7d09-4450-bd1e-4cb116e283d5

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
@@ -216,7 +216,7 @@ class BackfillShowAction @Inject constructor(
             }
           }
 
-          Card {
+          FullWidthCard {
             // Partitions
             div("mx-auto max-w-7xl sm:px-6 lg:px-8") {
               h2("text-base font-semibold leading-6 text-gray-900") { +"""Partitions""" }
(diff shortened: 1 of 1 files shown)


jeffhwang-sq:
> LG
> 

> For super wide screen, you could likely make the max-width for partitions and events even wider to minimize any scrolling. Would be interesting to see what the examples with stacktraces in the events look like on the super wide.

Removed some additional paddings - Screenshot 2025-07-24 at 12 08 30 AM

service/src/main/kotlin/app/cash/backfila/ui/pages/BackfillShowAction.kt
```diff
@@ -323,9 +329,9 @@ class BackfillShowAction @Inject constructor(
 
         // Smart auto-reload section for Events (preserves expanded state)
         AutoReload(frameId = "backfill-$id-events") {
-          Card {
+          FullWidthCard {
             // Events
-            div {
+            div("mx-auto max-w-7xl sm:px-6 lg:px-8") {
```
mpawliszyn, line 334: Don't we want something like `max-w-max` ? We want the main window to take up the remainder but with some min size.

PR #506, opened 2025-08-01
[client-sqldelight-gradle-plugin] Update the generated methods and queries to be prefixed by the record source config name

M-Porter:
## Description

This PR comes about from me running into issues with how backfila + sqldelight interact with each other and the generated code not doing _quite_ what I'd want it to do.

The TL;DR of my issue can be boiled down to:
* I have a table `users` with a column `phone_number`.
* The type of `phone_number` is `String?`
* I am trying to add a new record source that uses `whereClause = "phone_number IS NOT NULL"` in the config
* During code generation, sqldelight sees that `phone_number` no longer maps to the `users` table as the type is now `String` (because `whereClause = "phone_number IS NOT NULL"` enforces that `phone_number` couldn't possibly be null)
    * Because of this sqldelight creates a new record type matching the query name. In the case of the generated code from the gradle plugin, this type would be `GetBatch`.
* While I _could_ use `GetBatch` as the type, this is not very intuitive or "useful" (you wouldn't call and apple an orange).

## Solution

This PR updates the code generation to prefix the `getBatch` method and query with the record source config name.

So given the following
```kotlin
addRecordSource(
    name = "usersWithPhoneNumbers",
    ...
    whereClause = "phone_number IS NOT NULL"
)
```
The generated query would be `usersWithPhoneNumbersGetBatch` and the record type would be `UsersWithPhoneNumbersGetBatch`.

Not only does this change cover the `where x is not null` case, it also covers cases where you specify specific columns in the `recordColumns` field.

## Alternatives

Because sqldelight will default the type to the query name if the result doesn't match the table view, an alternative would be to drop the `getBatch` suffix from the generated code.

So given the following
```kotlin
addRecordSource(
    name = "usersWithPhoneNumbers",
    ...
    whereClause = "phone_number IS NOT NULL"
)
```

The generated query would be `usersWithPhoneNumbers` and the record type would be `UsersWithPhoneNumbers`.

### DIFF
--- a/client-sqldelight-gradle-plugin/src/main/kotlin/app/cash/backfila/client/sqldelight/plugin/GenerateBackfilaRecordSourceQueriesTask.kt
+++ b/client-sqldelight-gradle-plugin/src/main/kotlin/app/cash/backfila/client/sqldelight/plugin/GenerateBackfilaRecordSourceQueriesTask.kt
@@ -84,7 +84,7 @@ abstract class GenerateBackfilaRecordSourceQueriesTask : DefaultTask() {
           ).addFunction(
             FunSpec.builder("selectAbsoluteRange")
               .returns(minMaxQueryType)
-              .addStatement("return database.%L.${lowerName}SelectAbsoluteRange { min, max -> %T(min, max) }", queriesFunctionName, minMaxType)
+              .addStatement("return database.%L.%LSelectAbsoluteRange { min, max -> %T(min, max) }", queriesFunctionName, lowerName, minMaxType)
               .addModifiers(OVERRIDE)
               .build(),
           ).addFunction(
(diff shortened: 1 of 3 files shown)


client-sqldelight-gradle-plugin/src/main/kotlin/app/cash/backfila/client/sqldelight/plugin/GenerateBackfilaRecordSourceQueriesTask.kt
```diff
@@ -159,7 +160,7 @@ abstract class GenerateBackfilaRecordSourceQueriesTask : DefaultTask() {
               .addParameter("start", keyType)
               .addParameter("end", keyType)
               .returns(recordQueryType)
-              .addStatement("return database.%L.getBatch(start, end)", queriesFunctionName)
+              .addStatement("return database.%L.%LGetBatch(start, end)", queriesFunctionName, lowerName)
```
M-Porter, line 163: Is this preferred to `%LGetBatch`?
```suggestion
              .addStatement("return database.%L.%L(start, end)", queriesFunctionName, "${lowerName}GetBatch")
```

M-Porter:
> Otherwise I'm likely to be lost.
Yeah that's a completely valid reason. I'm ok with keeping it as is now.

adrw (approved):
Looks like a reasonable change and solution to the problem to me.

PR #509, opened 2025-08-19
Make parameters classes easier to work with

fspinolo:
The key change is changing the type of
`BackfillRegistration.parametersClass` from `KClass<Any>` to `KClass<out Any>`. This is a clearer expression of the intent to allow backfill parameters to be of any type.

In the previous implementation, the type was `KClass<Any>` which actually means the class of the `Any` type. This required some gymnastics for the standard clients to be able to register themselves because they had to cast their parameters type to `Any` (usually by casting the backfill class to something like `SomeBackfill<Any, Any>` so that a parameters type of `Any` gets extracted).

With this change, a lot of these type gymnastics can be simplified. This also makes it much more straightforward to implement a custom client that uses a custom parameters type.

### DIFF
--- a/client-base/src/test/kotlin/app/cash/backfila/client/fixedset/FixedSetBackend.kt
+++ b/client-base/src/test/kotlin/app/cash/backfila/client/fixedset/FixedSetBackend.kt
@@ -18,12 +18,13 @@ import kotlin.reflect.KClass
 import kotlin.reflect.full.findAnnotation
 
 @Singleton
-class FixedSetBackend @Inject constructor(
+class FixedSetBackend
+@Inject
+constructor(
   private val injector: Injector,
   @ForFixedSetBackend private val backfills: MutableMap<String, KClass<out FixedSetBackfill<*>>>,
   private val datastore: FixedSetDatastore,
 ) : BackfillBackend {
-
   /** Creates Backfill instances. Each backfill ID gets a new Backfill instance. */
   private fun getBackfill(name: String): FixedSetBackfill<*>? {
     val backfillClass = backfills[name]
@@ -34,13 +35,16 @@ class FixedSetBackend @Inject constructor(
     }
   }
 
-  private fun <Param : Any> createOperator(
-    backfill: FixedSetBackfill<Param>,
-  ) = FixedSetBackfillOperator(
-    backfill = backfill,
-    datastore = datastore,
-    parametersOperator = BackfilaParametersOperator(parametersClass(backfill::class)),
-  )
+  @Suppress("UNCHECKED_CAST")
+  private fun <Param : Any> createOperator(backfill: FixedSetBackfill<Param>) =
+    FixedSetBackfillOperator(
+      backfill = backfill,
+      datastore = datastore,
+      parametersOperator =
+      BackfilaParametersOperator(
+        parametersClass(backfill::class),
+      ) as BackfilaParametersOperator<Param>,
+    )
 
   override fun create(backfillName: String): BackfillOperator? {
     val backfill = getBackfill(backfillName)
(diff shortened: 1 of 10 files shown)


client-base/src/test/kotlin/app/cash/backfila/client/fixedset/FixedSetBackend.kt
```diff
@@ -34,13 +35,16 @@ class FixedSetBackend @Inject constructor(
...
   }
 
-  private fun <Param : Any> createOperator(
-    backfill: FixedSetBackfill<Param>,
-  ) = FixedSetBackfillOperator(
-    backfill = backfill,
-    datastore = datastore,
-    parametersOperator = BackfilaParametersOperator(parametersClass(backfill::class)),
-  )
+  @Suppress("UNCHECKED_CAST")
```
mpawliszyn, line 38: Do we need this? I don't know that I like casting the operator, I'd rather cast the Param type exclusively.

client-base/src/test/kotlin/app/cash/backfila/client/fixedset/FixedSetBackend.kt
```diff
@@ -53,26 +57,26 @@ class FixedSetBackend @Inject constructor(
...
+          name = it.key,
+          description = it.value.findAnnotation<Description>()?.text,
+          parametersClass = parametersClass(it.value),
+          deleteBy = it.value.findAnnotation<DeleteBy>()?.parseDeleteByDate(),
+          unit = BackfillUnit.ITEMS.displayName,
+        )
+      }.toSet()
 
-  private fun <T : Any> parametersClass(backfillClass: KClass<out FixedSetBackfill<T>>): KClass<T> {
+  private fun parametersClass(backfillClass: KClass<out FixedSetBackfill<*>>): KClass<*> {
```
mpawliszyn, line 72: Yeah I don't think that removing this anchoring to the type is what I'm looking for.

I don't want parameters to be able to be any type for a particular backfill but only the parameterized type for that specific backfill.

client-jooq/src/main/kotlin/app/cash/backfila/client/jooq/internal/JooqBackend.kt
```diff
@@ -60,14 +63,14 @@ class JooqBackend @Inject constructor(
     }
   }
 
-  private fun <T : Any> parametersClass(backfillClass: KClass<out JooqBackfill<*, T>>): KClass<T> {
```
adrw, line 63: Does removing the generic mean the type isn't known at the callsite?

mpawliszyn (commented):
The constructor is actually important too:
https://github.com/cashapp/backfila/blob/master/client-base/src/main/kotlin/app/cash/backfila/client/spi/BackfilaParameters.kt#L67

Anyway it is an interesting point. I'm not entirely happy with the gymnastics and open to doing it another way as long as its type safe and keeps the complexity in the clients and out of the backfills themselves.

I'll think about it some more

client-jooq/src/main/kotlin/app/cash/backfila/client/jooq/internal/JooqBackend.kt
fspinolo: Before, we were always casting the type to `Any` so that it was "known" at the callsite. This was all to satisfy the type of `KClass<Any>` on `BackfillRegistration.parametersClass`. Using the type `KClass<Any>` meant you couldn't use any params type without casting it to `Any` first

client-base/src/test/kotlin/app/cash/backfila/client/fixedset/FixedSetBackend.kt
fspinolo: You still get that here. Try it for yourself, the types still have to match up

PR #511, opened 2025-08-26
Upgrading JDK, Gradle, Misk, Jooq gradle plugin and the com.gradleup.shadow plugin.

mpawliszyn:


jeffhwang-sq (approved):

bin/jar
```diff
@@ -1 +1 @@
-.openjdk@17.pkg
\ No newline at end of file
+.openjdk@21.pkg
```
staktrace, line 1: Generally it's not advisable to use channel packages if you want reproducible builds. Up to you though

staktrace (approved):
couple of comments but looks reasonable

gradle/libs.versions.toml
```diff
@@ -50,7 +50,7 @@ retrofitGuavaAdapter = { module = "com.squareup.retrofit2:adapter-guava", versio
 retrofitMock = { module = "com.squareup.retrofit2:retrofit-mock", version = "3.0.0" }
 retrofitMoshi = { module = "com.squareup.retrofit2:converter-moshi", version = "3.0.0" }
 retrofitWire = { module = "com.squareup.retrofit2:converter-wire", version = "3.0.0" }
-shadowJarPlugin = { module = "gradle.plugin.com.github.johnrengelman:shadow", version = "7.1.2" }
+shadowJarPlugin = { module = "com.gradleup.shadow:shadow-gradle-plugin", version = "9.0.2" }
```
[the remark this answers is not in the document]
mpawliszyn: Sounds good. I'll choose that one too.

bin/jar
mpawliszyn: Ah I think I understand what you are saying here, thanks for catching.

build.gradle.kts
```diff
@@ -42,12 +42,12 @@ subprojects {
     dependsOn("spotlessKotlinApply")
     compilerOptions {
       jvmTarget = JvmTarget.JVM_11
+      freeCompilerArgs.add("-Xjdk-release=11")
```
mpawliszyn, line 45: @staktrace Is there a reason why you did it this way?

I'm running into trouble with my gradle plugin now
`e: '-Xjdk-release=11' option conflicts with '-jvm-target 17'. Please remove the '-jvm-target' option`

mpawliszyn: Yeah the gradle plugin sets the target to 17 because it needs it for SqlDelight integration. Plus the artifact is a gradle plugin not a java library.

mpawliszyn: This here:
https://github.com/cashapp/backfila/blob/master/client-sqldelight-gradle-plugin/build.gradle.kts#L47-L55

PR #512, opened 2025-08-26
add Backfill Search Dropdown Implementation

jeffhwang-sq:
This change was added to the legacy UI as an additional functionality. We want the new UI to have this functionality as well.

https://github.com/user-attachments/assets/ba9ede4c-bbf8-4afc-807a-22efe0c62583

https://github.com/user-attachments/assets/a420c368-c8a1-45a2-8127-d9637bffee7c


mpawliszyn:
Does typing it out work too? Some folks have a lot of backfills.

jeffhwang-sq:
> Does typing it out work too? Some folks have a lot of backfills.

Yes, it does. Will post a video

mpawliszyn (approved):

PR #513, opened 2025-08-28
Fix search filter form submission for services with default variants

jeffhwang-sq:
Fix search filter form submission for services with default variants


service/src/main/kotlin/app/cash/backfila/ui/components/BackfillSearchForm.kt
```diff
@@ -21,6 +21,7 @@ fun TagConsumer<*>.BackfillSearchForm(
     form(classes = "flex flex-wrap gap-4 items-end") {
       method = kotlinx.html.FormMethod.get
       attributes["data-turbo-frame"] = "_top"
+      action = "/services/$serviceName/" + if (variantName != "default") variantName else ""
```
mpawliszyn, line 24: else "" or "default" ?

PR #514, opened 2025-08-28
Add service info modal to display service metadata

jeffhwang-sq:
Adding a new "Info" button and modal popup to the service page that displays detailed service metadata including connector information, registration timestamps, and configuration details.

https://github.com/user-attachments/assets/5b8ba2df-bd32-4a22-a429-2d7632a635f6

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/UiModule.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/UiModule.kt
@@ -10,6 +10,7 @@ import app.cash.backfila.ui.pages.BackfillIndexAction
 import app.cash.backfila.ui.pages.BackfillShowAction
 import app.cash.backfila.ui.pages.IndexAction
 import app.cash.backfila.ui.pages.ServiceIndexAction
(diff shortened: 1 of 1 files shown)


service/src/main/kotlin/app/cash/backfila/ui/pages/ServiceInfoAction.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    val variant = variantOrBlank.orEmpty().ifBlank { "default" }
+
+    try {
+      val serviceDetails = getServiceDetailsAction.getServiceDetails(service, variant)
+      val label = if (variant == "default") service else "$service ($variant)"
+
+      val htmlResponseBody = dashboardPageLayout.newBuilder()
+        .title("$label Info | Backfila")
+        .breadcrumbLinks(
```
mpawliszyn, line 53: I wonder if we can apply these in a way that isn't so much copy and paste?

mpawliszyn (approved):
A great page!

PR #515, opened 2025-08-28
Added a development service without explicit variants

jeffhwang-sq:
Added a development service without explicit variants for testing and fixed the search filter functionality for services using default variants.

https://github.com/user-attachments/assets/9427f0a3-a488-430f-b54f-9a23b803febf

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/components/BackfillSearchForm.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/components/BackfillSearchForm.kt
@@ -21,7 +21,6 @@ fun TagConsumer<*>.BackfillSearchForm(
     form(classes = "flex flex-wrap gap-4 items-end") {
       method = kotlinx.html.FormMethod.get
       attributes["data-turbo-frame"] = "_top"
-      action = "/services/$serviceName/" + if (variantName != "default") variantName else ""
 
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/app/cash/backfila/ui/components/BackfillSearchForm.kt
```diff
@@ -21,7 +21,6 @@ fun TagConsumer<*>.BackfillSearchForm(
     form(classes = "flex flex-wrap gap-4 items-end") {
       method = kotlinx.html.FormMethod.get
       attributes["data-turbo-frame"] = "_top"
-      action = "/services/$serviceName/" + if (variantName != "default") variantName else ""
```
adrw, line 24: Is this unused? I assume removing this will break the form?

mpawliszyn (changes requested):
We don't want to confuse McDees and FineDining. Feel free to add another client but it should be named differently. McDees has variants and FineDining does not. But we can have more that's fine.

jeffhwang-sq:
> We don't want to confuse McDees and FineDining. Feel free to add another client but it should be named differently. McDees has variants and FineDining does not. But we can have more that's fine.

Okay I ended up removing the new service and just using the finedining to correctly test it out.

mpawliszyn (approved):

PR #518, opened 2025-09-02
revamp search funcationality

jeffhwang-sq:
Replaced HTML5 datalist search with a robust autocomplete system using Stimulus controllers, providing better UX for finding backfills and users. 
- Real-time autocomplete for both backfill names and users
- Click to select with hover states
- Improved clear filters

Removed GetBackfillNamesAction and added SearchBackfillNamesAction, SearchUsersAction
Checked regular autoloading works fine when backfill runs.

https://github.com/user-attachments/assets/bee04b22-1876-46a0-8205-0a8aeebb3322

Add keyboard support

https://github.com/user-attachments/assets/4dcbcefc-d976-4cd8-a755-09c8971c28f3


service/src/main/kotlin/app/cash/backfila/dashboard/SearchUsersAction.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    return transacter.transaction { session ->
+      val dbService = queryFactory.newQuery<ServiceQuery>()
+        .registryName(service)
+        .variant(variant)
+        .uniqueResult(session) ?: throw BadRequestException("`$service`-`$variant` doesn't exist")
+
+      // Get unique users who have created backfills for this service
+      val users = session.hibernateSession.createQuery(
+        """
```
mpawliszyn, line 44: I think we want to use a Query class.

jeffhwang-sq: The reason why we have the raw hibernate query is because misk query framwork does not support Like operation, and we would need to query all the backfills if we ues the query class.

adrw:
Does the new search UI work with keyboard arrow keys and then [enter] to select from the available responses?

jeffhwang-sq:
> Does the new search UI work with keyboard arrow keys and then [enter] to select from the available responses?

Added and posted in the test

service/src/main/resources/web/static/js/autocomplete_controller.js
```diff
@@ -1,25 +1,50 @@
...
- * </div>
- *
- * The server is expected to send back snippets as such:
- *
- * <li class="list-group-item" role="option" data-autocomplete-value="1">Blackbird</li>
- * <li class="list-group-item" role="option" data-autocomplete-value="2">Bluebird</li>
- * <li class="list-group-item" role="option" data-autocomplete-value="3">Mockingbird</li>
- *
- */
-Stimulus.register("autocomplete", Autocomplete)
```
adrw, line 24: What was the main motivation for doing a custom controller instead of reusing the Autocomplete library?

service/src/main/kotlin/app/cash/backfila/dashboard/SearchBackfillNamesAction.kt
```diff
@@ -4,32 +4,36 @@ import app.cash.backfila.service.persistence.BackfilaDb
...
 import misk.web.ResponseContentType
 import misk.web.actions.WebAction
 import misk.web.mediatype.MediaTypes
 
-data class GetBackfillNamesResponse(
-  val backfill_names: List<String>,
-)
-
-class GetBackfillNamesAction @Inject constructor(
+class SearchBackfillNamesAction @Inject constructor(
```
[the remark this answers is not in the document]
jeffhwang-sq: Got it. WIll leave the json around.

jeffhwang-sq:
Updated the pr to use misk sql query with newly added Like operator

PR #548, opened 2026-03-25
URL-encode service names to fix slash handling in URLs

borisilic-ap:
## Summary
- URL-encode service names when building URLs so `/` in names like `ki/ki` becomes `%2F`, preventing the Misk router from splitting the name into separate path segments
- Applied the fix to `path()` companion methods in `ServiceShowAction`, `ServiceInfoAction`, `BackfillCreateServiceIndexAction`, and `BackfillCreateAction`
- Fixed sidebar link construction in `DashboardPageLayout` which also used raw string interpolation for service names
- Added `ServiceShowActionTest` with tests verifying that slashes in service names are properly encoded

## Test plan
- [x] `ServiceShowActionTest` passes: `service name with slash should not produce extra path segments` verifies URL encoding
- [x] `ServiceShowActionTest` passes: `service with slash in name should be reachable` verifies end-to-end service registration and lookup with slashed names


borisilic-ap:
not needed. solved in other PR.

PR #550, opened 2026-04-07
Fix completed backfills not showing 100% progress

petera-block:
## Summary
- Backfills with state `COMPLETE` could show less than 100% progress when the backfilled count didn't exactly match the computed count
- Clamp `ProgressBar` percentage to `0–100%` via `.coerceIn(0.0, 100.0)`
- Override progress to 100% on both the backfill detail page and the backfills list table when state is `COMPLETE`

## Test plan
- [x] `./gradlew :service:build` passes
- [ ] Verify on a completed backfill (e.g. [#38168](https://backfila.sqprod.co/backfills/38168)) that progress now shows 100%
- [ ] Verify in-progress and paused backfills still show accurate partial progress

### DIFF
--- a/service/src/main/kotlin/app/cash/backfila/ui/components/BackfillsTable.kt
+++ b/service/src/main/kotlin/app/cash/backfila/ui/components/BackfillsTable.kt
@@ -1,6 +1,7 @@
 package app.cash.backfila.ui.components
 
 import app.cash.backfila.dashboard.UiBackfillRun
+import app.cash.backfila.service.persistence.BackfillState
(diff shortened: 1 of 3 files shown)


jessejacksonafterpay:
@petera-block I took a slightly different approach in this PR to use decimals on the progress bar rather than rounding.
https://github.com/cashapp/backfila/pull/551

mpawliszyn:
So I'm not really a fan of pretending that it went to 100% of what it calculated when it didn't. there is something to be learned by noticing that it is "done" at 98%. Since this means that there was a different count at calculation vs execution time.

PR #552, opened 2026-05-11
Filter partitions by state and sort by progress

TheSentinel454:
## Summary
- On the backfill detail page, the partitions table can be long and visually uniform. Adds a **state filter** (All / RUNNING / PAUSED / COMPLETE / CANCELLED) and **click-to-sort** on the Progress (%) header (cycles none → desc → asc → none).
- Both operate client-side via a small new Stimulus controller. State is persisted in the URL query string (`?state=RUNNING&sort=progress_desc`) so it survives the 10s auto-reload tick and is shareable.
- Partitions still computing progress are always sorted to the bottom regardless of sort direction.
- Includes `dev/seed_partitions.sql` for populating an existing backfill with mixed-state synthetic partitions to test the new UI locally.

## Implementation notes
- Pure additive change: row-level `data-partition-state` / `data-progress-pct` attributes are emitted server-side; the controller reads them, hides non-matching rows via the `hidden` class, and reorders via `tbody.appendChild`.
- No changes to `auto_reload_controller.js`. Because the controller's element lives inside the partitions turbo-frame, every auto-reload swap triggers a fresh `connect()` which re-reads the URL and re-applies filter + sort — no custom cross-frame events needed.

## Test plan
- [x] Start the dev service + a client (e.g. McDees), create + start a backfill.
- [x] Run `mysql -u root backfila_development --init-command="SET @run_id := <ID>" < dev/seed_partitions.sql` to seed 10 partitions in mixed states.
- [x] Reload the backfill detail page; verify all 10 rows render with the new filter dropdown and sortable Progress (%) header.
- [x] Change filter to `RUNNING` — only the 5 running rows remain.
- [x] Click Progress (%) — rows sort desc; computing/precomputing rows go to bottom. Click again for asc.
- [x] Wait through a 10s auto-reload tick; verify filter + sort persist.
- [x] Copy the URL and open in a new tab; verify filter + sort are restored from query string.

https://github.com/user-attachments/assets/2b356c34-6d10-439b-ad91-eda248a974e7

### DIFF
--- a/dev/seed_partitions.sql
+++ b/dev/seed_partitions.sql
@@ -1,0 +1,47 @@
+-- Dev-only seed for the partitions table UI (testing filter + sort).
+--
+-- Replaces an existing backfill_run's partitions with 10 synthetic rows in
+-- mixed states (RUNNING/PAUSED/COMPLETE/CANCELLED) and varied progress.
+--
+-- Usage:
+--   1. In the Backfila UI, create + start any backfill (e.g. BurgerFlippingBackfill).
+--   2. Find its backfill_run_id (visible in the backfill detail page URL, or:
+--        mysql -u root backfila_development -e \
+--          "SELECT id, state FROM backfill_runs ORDER BY id DESC LIMIT 5;"
+--   3. Run this script, passing the id via --init-command:
+--        mysql -u root backfila_development \
+--          --init-command="SET @run_id := <BACKFILL_RUN_ID>" \
+--          < dev/seed_partitions.sql
+--   4. Reload the backfill detail page.
+--
+-- Notes:
+--   - lease_expires_at is set near the MySQL TIMESTAMP cap (2038-01-19) to keep
+--     the real runner from leasing these rows. Values past 2038 silently get
+--     stored as zero — which Hibernate then rejects as "Zero date value prohibited".
+--   - If the parent backfill is RUNNING, the runner will skip these synthetic
+--     partitions (lease guard). If you want a truly static view, pause it.
+
+SELECT CONCAT('Seeding 10 synthetic partitions for backfill_run_id = ', @run_id) AS status;
+
+DELETE FROM run_partitions WHERE backfill_run_id = @run_id;
+
+INSERT INTO run_partitions
+  (backfill_run_id, partition_name, run_state, lease_expires_at,
+   pkey_range_start, pkey_range_end, pkey_cursor,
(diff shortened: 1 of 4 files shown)


dev/seed_partitions.sql
```diff
@@ -0,0 +1,47 @@
+-- Dev-only seed for the partitions table UI (testing filter + sort).
```
mpawliszyn, line 1: I don't love how brittle this is but I don't have a better solution right now.

service/src/main/resources/web/static/js/partitions_table_controller.js
```diff
@@ -0,0 +1,78 @@
+import { Application, Controller } from "../cache/stimulus/3.1.0/stimulus.min.js";
+window.Stimulus = Application.start();
+
+Stimulus.register("partitions-table", class extends Controller {
+  static targets = ["filter", "tbody", "sortIndicator"];
+
+  connect() {
+    const params = new URLSearchParams(window.location.search);
```
mpawliszyn, line 8: @adrw WDYT?

mpawliszyn (approved):
I'd appreciate @adrw 's review first. But if he doesn't have time, its a small enough change.

dev/seed_partitions.sql
TheSentinel454: Yeah, I'm open to just not committing this.  I used this locally to just add some test data and agree this isn't the best long term approach.  It seemed like a more niche usage for me.

TheSentinel454:
Going to go ahead and merge. @adrw Let me know if you have any concerns and I can address them in a later PR.

service/src/main/resources/web/static/js/partitions_table_controller.js
adrw: Looks reasonable, I do like the permalink capability here.

dev/seed_partitions.sql
adrw: @mpawliszyn do we have a seed data setup already or we usually just run the test service alongside Backfila and manually create a few backfills through the UI for manual QA?

mpawliszyn: Just manually at the moment. I mostly don't like how tied this is to the current state of the data model. I'd rather some type of script to walk the system through a few states to produce these situations "naturally" instead. But that is a lot of yak shaving for this. 

We will end up playing whack a mole with this though.

PR #556, opened 2026-05-27
Update dokka to 2.x and setup a multi-module versioned setup.

escardin:
It's not any worse than what exists, but at least it's not complaining anymore. I think some small tweaks could get it all the way to working.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,0 +1,13 @@
+plugins {
+  `kotlin-dsl`
+}
+
+repositories {
+  mavenCentral()
+  gradlePluginPortal()
+}
+
+dependencies {
+  implementation(libs.dokkaGradlePlugin)
+  implementation(libs.kotlinGradlePlugin)
+}
--- a/.github/workflows/prepare_mkdocs.sh
+++ b/.github/workflows/prepare_mkdocs.sh
@@ -9,7 +9,7 @@
 set -ex
 
 # Generate the API docs
-gradle dokkaGfm
+gradle :docs:dokkaGenerateMarkdown
 
 # Dokka filenames like `-http-url/index.md` don't work well with MkDocs <title> tags.
 # Assign metadata to the file's first Markdown heading.
@@ -25,7 +25,7 @@ title_markdown_file() {
 }
 
 set +x
-for MARKDOWN_FILE in $(find docs/0.x/ -name '*.md'); do
+for MARKDOWN_FILE in $(find docs/build/dokka/markdown -name '*.md'); do
   echo $MARKDOWN_FILE
   title_markdown_file $MARKDOWN_FILE
 done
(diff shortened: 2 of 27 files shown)


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
...
+  `kotlin-dsl`
+}
+
+repositories {
+  mavenCentral()
+  gradlePluginPortal()
+}
+
+dependencies {
+//  implementation(libs.mavenPublishGradlePlugin)
```
lkerford: nit: remove these comment out deps

lkerford:
I don't see any change in the PR which would cause the test not to run. I don't think its related to this change but the test check isn't completed on this PR. It could be as simple as trying to re-triggering the github action by rebasing the PR

lkerford (approved):

buildSrc/build.gradle.kts
escardin: Oh right, I keep wanting to make those lines work and forgetting. It looks like the build is failing too, probably for similar reasons.

adrw:
Can you include the original error logs or motivation for this PR?

What is the motivation for the `multi-module versioned`? I'm not immediately seeing why we need the separate buildSrc build gradle with explicit version catalog setup when the version catalog seemed to be working fine before?

escardin:
> Can you include the original error logs or motivation for this PR?
> 
> What is the motivation for the `multi-module versioned`? I'm not immediately seeing why we need the separate buildSrc build gradle with explicit version catalog setup when the version catalog seemed to be working fine before?

Actually, the docs website doesn't work (and doesn't truly work after this either). I wanted to get rid of the Dokka 2.x warnings, and the way I changed it is the recommended way to go about doing what we were doing according to the docs. That is, publishing versioned docs is supported in dokka, but we were doing some weird hack. Using a convention plugin is preferred over the anti-pattern of `subprojects{}` and `allprojects{}` blocks.

We also had notes that we were doing a post config to update paths because the publishing plugin was mangling things, but doing this way everything works as things were designed to. There are some extra things to do to fully fix the docs site, but this is a step in the right direction.

escardin:
This is the warning when trying to use the V1 mode of the plugin:
```
warning: Dokka Gradle plugin V1 is deprecated
    
    Dokka Gradle plugin V1 is deprecated, and will be removed in Dokka version 2.1.0
    Please migrate to Dokka Gradle plugin V2. This will require updating your project.
    To learn about migrating read the migration guide https://kotl.in/dokka-gradle-migration
    
    To start migrating to Dokka Gradle plugin V2 add
        org.jetbrains.dokka.experimental.gradle.pluginMode=V2EnabledWithHelpers
    into your project's `gradle.properties` file.
    
    We would appreciate your feedback!
     - Please report any feedback or problems https://kotl.in/dokka-issues
     - Chat with the community visit #dokka in https://kotlinlang.slack.com/ (To sign up visit https://kotl.in/slack)
```

PR #570, opened 2026-05-27
Add support for using Guice 6 or 7

escardin:
Plausibly a breaking change in RealBackfillModule, but should be an easy fix. I'm happy to have a jakarta provider, but it seems odd to have to manually provide a config? An instance should be lots.

I am not upgrading the deps to Guice 7, but I did run the tests against it and they passed.

Nav uses Guice 7, and misk supports it, even if it doesn't demand it. Because Backfila doesn't provide enough to just be deployed standalone, and we need a wrapper around it to deploy it, it'd be real helpful if we can stay on Guice 7.


client-base/src/main/kotlin/app/cash/backfila/client/RealBackfillModule.kt
```diff
@@ -6,37 +6,27 @@ import app.cash.backfila.client.internal.RealBackfilaClient
 import app.cash.backfila.client.spi.BackfillBackend
 import com.google.inject.AbstractModule
 import com.google.inject.multibindings.Multibinder
-import javax.inject.Provider
 import kotlin.reflect.KClass
 
 class RealBackfillModule(
-  private val configProvider: Provider<BackfilaClientConfig>,
```
escardin, line 13: The api changes in this file are a breaking change. I don't know if RealBackfillModule is ever constructed outside of MiskBackfillModule in the Faire or Block codebases, but please be aware of this.

I'm happy to keep it as a provider, so long as it can be a jakarta provider. That can be wrapped with Providers.guicify or I can write an `@Provides` for it.

It honestly feels like a code smell to have a module depend on a provider though.

client-jooq/src/test/kotlin/app/cash/backfila/client/jooq/config/ClientJooqTestingModule.kt
```diff
@@ -32,16 +30,16 @@ import org.jooq.impl.DefaultTransactionProvider
 import org.jooq.tools.JooqLogger
 
 class ClientJooqTestingModule : KAbstractModule() {
```
escardin, line 32: This is just used for tests, so I'm not super worried about the changes being breaking. Using `@Provides` is still much cleaner than all the hoops that were jumped through for a binding.

escardin (commented):
The vast majority of the changes are import swaps that are fully compatible with both guice 6 & 7. I called out the few spots with changes that aren't just import changes,

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfillModule.kt
ThePumpingLemma: You can use `com.google.inject.Provider` for Guice 6 + 7 compat.

escardin: It still results in a breaking change if it's used directly. If that's the case, may as well deal with the code smell.

At the end of the day, passing in a provider only to bind it is just weird and awkward. It should be bound externally and injected into the module, or the instance passed. Passing the config instance is much more standard and is what backfila itself is doing.

lkerford (approved):

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfillModule.kt
adrw: I'd split this out into a separate PR since this does change the API, that way this PR can be strictly updating the imports for Guice 6/7. This PR is ready to stamp once that's done.

escardin: So, do it as a two step PR, of add the new, and then remove the old in a followup?

adrw: Yeah to minimize the API changes in case there are any breakages. The jakartha import switch should be uneventful but these other API changes could be an issue, and ideally if there is any issue, the jakartha change can stick and doesn't need to be reverted.

adrw (approved):
LG

client-base/src/main/kotlin/app/cash/backfila/client/RealBackfillModule.kt
```diff
@@ -29,11 +35,23 @@ class RealBackfillModule(
...
+   */
+  @Suppress("unused")
+  @Deprecated("Providing an instance is preferred over a provider", replaceWith = ReplaceWith("RealBackfillModule(config,loggingSetupProvider)"))
   constructor(
-    configProvider: Provider<BackfilaClientConfig>,
+    configProvider: JakartaProvider<BackfilaClientConfig>,
     loggingSetupProvider: Class<out BackfilaClientLoggingSetupProvider> =
       BackfilaClientNoLoggingSetupProvider::class.java,
-  ) : this(configProvider, loggingSetupProvider.kotlin)
+  ) : this(Providers.guicify(configProvider), loggingSetupProvider.kotlin)
```
keeferrourke, line 76: Please use the com.google.inject.Provider. You don't need this `guicify` stuff, and you shouldn't need to deprecate these constructors either. Guice 6/7 both work with the google Provider, and is more consistent with Misk.

escardin: The issue is everything here is effectively backwards, and that's why I went with the breaking change in the first place. Passing a `Provider` into a module is _wrong_. If you need it to be a provider, then `@Inject` the module, or use a `@Provides` method. But neither of those make sense, because this is where we are binding the instance in the first place.

To the best of my knowledge, these are compile errors:
```kotlin
val a: com.google.inject.Provider = jakarta.inject.Provider{"a"}
val b: com.google.inject.Provider = javax.inject.Provider{"b"}
```

Versus this, which is valid.
```kotlin
val a: jakarta.inject.Provider = com.google.inject.Provider{"a"}
val b: javax.inject.Provider = com.google.inject.Provider{"b"}
```

In order for this to be a non-breaking change, we must provide an API for AT LEAST the existing `javax.inject.Provider`. To be maximally compatible, I added `jakarta.inject.Provider` and `com.google.inject.Provider`, so that folks can use whatever they please. That said, passing ANY provider into a module constructor is a mistake, so I marked ALL of them deprecated, with the preferred replacement of passing the instance, which is what backfila actually does, with the goal of dropping the deprecated ones in a followup.

The real problem all this work is spent fixing is actually the line:
```kotlin
bind(BackfilaClientConfig::class.java).toProvider(configProvider)
```
It's a compile error in Guice 6 to pass a jakarta provider when it expects javax, and vice versa for Guice 7. 

[`Providers.guicify()`](https://github.com/google/guice/blob/55a2b68ebe445b8dca3795bd3cdfc5c09d566e74/core/src/com/google/inject/util/Providers.java#L108-L115) exists specifically to resolve this scenario, and is specifically called out as the solution in the [Guice 6 release notes](https://github.com/google/guice/wiki/Guice600):

> Guice 6.0 can help with incremental migrations to the jakarta.inject namespace, by incrementally replacing javax.inject references to jakarta.inject. This works everywhere, except for code where a jakarta Provider is passed to bind(..).toProvider. Any one of the following options can be used to workaround the toProvider issue.
> 
> Update to Guice 7.0 instead. Guice 7.0 fully supports jakarta.inject (but does not have support for javax.inject anymore).
> Change the provider implementation to instead implement com.google.inject.Provider (which works in both 6.0 & 7.0).
> Change the provider implementations to instead be @Provides methods, or add a wrapper @Provides method that delegates to the actual provider.
> Call bind(..).toProvider(Providers.guicify(myJakartaProvider)). The [guicify](https://google.github.io/guice/api-docs/6.0.0/javadoc/com/google/inject/util/Providers.html#guicify(jakarta.inject.Provider)) method will adapt a jakarta provider instance to a com.google.inject.Provider. This only works if binding a Provider instance. It will not work for binding provider classes, type literals, or keys. If using this approach, it's best to revert the guicify calls when later updating to Guice 7.0, because they otherwise mask the actual provider implementation (which can be accessed through the Guice SPI's [ProviderInstanceBinding.getUserSuppliedProvider](https://google.github.io/guice/api-docs/6.0.0/javadoc/com/google/inject/spi/ProviderInstanceBinding.html#getUserSuppliedProvider())).

keeferrourke: Ok I see your point. I was concerned about introducing immediately deprecated constructors, but if they are transient, this probably is the most compatible upgrade path. I totally misread that the javax.Provider was the existing ctor before this diff.

keeferrourke (approved):
LGTM

oldergod (approved):
Are we happy breaking the Kotlin callers?

We're dropping the old public constructor shape `javax.inject.Provider<BackfilaClientConfig>, KClass<...>`. Existing Kotlin callers using a Provider and SomeLoggingProvider::class will no longer compile, and already-compiled consumers can hit NoSuchMethodError.

keeferrourke:
>Are we happy breaking the Kotlin callers?

No, we are not. It should be possible to make this PR backward and forward compatible. I can migrate callers internally before this PR lands, but I will probably miss stuff.

oldergod:
- [ ] rebase on #585 and make sure we have no breaking changes before merging

adrw (approved):
LG

PR #571, opened 2026-05-27
Fixes to support Postgres as a backend DB

escardin:
This is not complete support for postgres. I am deliberately not including the migrated schema, or any wiring. These are the minimal changes I need to get it running with my own misk application wrapper.

I'll add more of a build out for postgres once I'm more confident it works and won't mysteriously fall over, as well as ensuring the tests work with both postgres and mysql at a later date.

### DIFF
--- a/client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/DbActiveCoupon.kt
+++ b/client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/DbActiveCoupon.kt
@@ -4,6 +4,7 @@ import java.time.Instant
 import javax.persistence.Column
 import javax.persistence.Entity
 import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
 import javax.persistence.Table
 import misk.hibernate.Constraint
 import misk.hibernate.DbUnsharded
@@ -17,7 +18,7 @@ import org.hibernate.annotations.Where
 @Where(clause = "expired_at IS NULL")
 class DbActiveCoupon() : DbUnsharded<DbActiveCoupon> {
   @javax.persistence.Id
-  @GeneratedValue
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
   override lateinit var id: Id<DbActiveCoupon>
 
   @Column
--- a/client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/DbMenu.kt
+++ b/client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/DbMenu.kt
@@ -3,6 +3,7 @@ package app.cash.backfila.client.misk
 import javax.persistence.Column
 import javax.persistence.Entity
 import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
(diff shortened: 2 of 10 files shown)


client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/DbActiveCoupon.kt
```diff
@@ -17,7 +18,7 @@ import org.hibernate.annotations.Where
 @Where(clause = "expired_at IS NULL")
 class DbActiveCoupon() : DbUnsharded<DbActiveCoupon> {
   @javax.persistence.Id
-  @GeneratedValue
```
escardin, line 20: Mysql uses IDENTITY by default, but postgres uses SEQUENCE. This is a problem because if you're using a generated id column, you actually want to set it to IDENTITY.

service/src/main/kotlin/app/cash/backfila/dashboard/GetBackfillRunsAction.kt
```diff
@@ -154,7 +154,7 @@ class GetBackfillRunsAction @Inject constructor(
     val list = session.hibernateSession.createQuery(
       """
         select backfill_run_id,
-          sum(precomputing_done) = sum(1),
```
escardin, line 157: Mysql doesn't support `bool_and` and postgres won't coerce bools so `sum()` or `bit_and()` will work without an explicit cast. Casting like that is afaik DB specific, and this is the best solution I could come up with to support both.

service/src/main/kotlin/app/cash/backfila/service/persistence/DbEventLog.kt
```diff
@@ -38,7 +39,7 @@ class DbEventLog() : DbUnsharded<DbEventLog>, DbTimestampedEntity {
   var partition_id: Id<DbRunPartition>? = null
 
   /** User that triggered the event, null when the event was triggered by the service. */
-  @Column
+  @Column(name = "`user`")
```
escardin, line 42: user is a reserved keyword in postgres. Mysql doesn't care about the escaping, and backticks are converted properly by hibernate for both DBs.

lkerford (approved):

client-misk-hibernate/src/test/kotlin/app/cash/backfila/client/misk/DbMenu.kt
```diff
@@ -14,7 +15,7 @@ import misk.hibernate.Query
 @Table(name = "menu")
 class DbMenu() : DbUnsharded<DbMenu> {
   @javax.persistence.Id
-  @GeneratedValue
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
```
adrw, line 18: Is this a change or just explicitly setting the default?

escardin: It's explicitly a change from AUTO to IDENTITY, but does not result in a change for Mysql based hibernate.

adrw (approved):
LG

PR #574, opened 2026-05-28
Update maven publishing plugin and switch to using a convention

escardin:
Should be a lot easer to work with and make changes to now, as we don't need to deal with things being applied indiscriminately to the project.

This is based on top of the dokka branch as it relies on some of what was done there to work.

I did this anticipating it would solve some issues with how we publish the service, but in the end it wasn't strictly necessary. I still prefer it, and it's more gradle-y.

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -1,12 +1,12 @@
 plugins {
   `java-platform`
-  id("com.vanniktech.maven.publish.base")
+  id("publishing-convention")
 }
 
 dependencies {
   constraints {
     project.rootProject.subprojects.forEach { subproject ->
-      if (subproject.name != "bom") {
+      if (subproject.name !in listOf("bom", "client-sqldelight-test")) {
         api(subproject)
       }
     }
--- a/buildSrc/src/main/kotlin/kotlin-publishing-convention.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-publishing-convention.gradle.kts
@@ -1,0 +1,14 @@
+import com.vanniktech.maven.publish.JavadocJar
+import com.vanniktech.maven.publish.KotlinJvm
+
+plugins {
+  id("publishing-convention")
+  id("dokka-convention")
+  kotlin("jvm")
+}
+mavenPublishing {
+
+  configure(
+    KotlinJvm(javadocJar = JavadocJar.Dokka("dokkaGeneratePublicationMarkdown"))
+  )
+}
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -21,8 +21,8 @@ git clone $REPO $DIR
 cd $DIR
 
 # Generate the API docs
-GRADLE_TASKS=$(ls -d misk*/ | cut -f1 -d'/' | awk '{ printf ":%s:dokkaGfm ", $1 }')
-gradle dokkaGfm
+GRADLE_TASKS=$(ls -d misk*/ | cut -f1 -d'/' | awk '{ printf ":%s:dokkaGenerateMarkdown ", $1 }')
+gradle dokkaGenerateMarkdown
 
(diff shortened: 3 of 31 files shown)


bom/build.gradle.kts
```diff
@@ -1,12 +1,12 @@
...
   `java-platform`
-  id("com.vanniktech.maven.publish.base")
+  id("publishing-convention")
 }
 
 dependencies {
   constraints {
     project.rootProject.subprojects.forEach { subproject ->
-      if (subproject.name != "bom") {
+      if (subproject.name !in listOf("bom", "client-sqldelight-test")) {
```
escardin, line 9: I'm adding an exclusion here because we don't actually publish the test jar (and shouldn't be). This makes the pom more accurate

buildSrc/src/main/kotlin/kotlin-publishing-convention.gradle.kts
```diff
@@ -0,0 +1,14 @@
+import com.vanniktech.maven.publish.JavadocJar
```
escardin, line 1: This is a separate convention because the bom subproject can't have the kotlin plugin configured even with `apply(false)`. It makes things clearer, even if somewhat less explicit than copying the same block everywhere.

buildSrc/src/main/kotlin/kotlin-publishing-convention.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+
+plugins {
+  id("publishing-convention")
+  id("dokka-convention")
+  kotlin("jvm")
+}
+mavenPublishing {
+
+  configure(
+    KotlinJvm(javadocJar = JavadocJar.Dokka("dokkaGeneratePublicationMarkdown"))
```
escardin, line 12: This is the correct target to use when pulling docs with dokka 2.x and it got missed before.

docs/build.gradle.kts
```diff
@@ -4,24 +4,10 @@ plugins {
...
-  dokka(project(":client-sqldelight-test"))
-  dokka(project(":client-static"))
-  dokka(project(":client-testing"))
-  dokka(project(":service"))
-  dokka(project(":service-self-backfill"))
-  
+  project.rootProject.subprojects.forEach { subproject ->
+    if (subproject.name !in listOf("client-sqldelight-test"))
+      dokka(subproject)
+  }
```
escardin, line 10: I saw this used in the bom subproject and I liked that it'd save folks effort if adding more subprojects to keep the docs working well.

deploy_website.sh
```diff
@@ -21,8 +21,8 @@ git clone $REPO $DIR
 cd $DIR
 
 # Generate the API docs
-GRADLE_TASKS=$(ls -d misk*/ | cut -f1 -d'/' | awk '{ printf ":%s:dokkaGfm ", $1 }')
-gradle dokkaGfm
+GRADLE_TASKS=$(ls -d misk*/ | cut -f1 -d'/' | awk '{ printf ":%s:dokkaGenerateMarkdown ", $1 }')
+gradle dokkaGenerateMarkdown
```
escardin, line 25: I guess this got missed in the dokka pr...

adrw:
A little bit hesitant to land this since it means Backfila drifts from the publishing patterns of other Block OSS repos / Misk which makes maintenance a bit more difficult. Seems like a reasonable change but want to think about it a bit more.

escardin:
> A little bit hesitant to land this since it means Backfila drifts from the publishing patterns of other Block OSS repos / Misk which makes maintenance a bit more difficult. Seems like a reasonable change but want to think about it a bit more.

Yeah. I'm okay with either not landing it, or doing at least the same thing for misk if that's reasonable?

I didn't end up needing it for what I was trying to accomplish (fixing publishing the unshaded jar), so it can be closed or deferred.

PR #575, opened 2026-05-28
Publish the unshaded jar without a classifier

escardin:
Maven doesn't know what to do with non-standard classifiers and they don't get listed in the pom. Because of this, the javadoc and source jars don't get linked to the unshaded jar (they might if they all had unshaded on them).

This makes trying to debug backfila a huge pain without running it in a very different way, and really doesn't accomplish anything in the end.

### DIFF
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -1,6 +1,5 @@
-import com.vanniktech.maven.publish.JavadocJar.Dokka
-import com.vanniktech.maven.publish.KotlinJvm
-import com.vanniktech.maven.publish.MavenPublishBaseExtension
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import com.vanniktech.maven.publish.tasks.JavadocJar
 
 plugins {
   id("com.gradleup.shadow")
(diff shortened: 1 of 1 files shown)


service/build.gradle.kts
```diff
@@ -96,10 +98,9 @@ val jar by tasks.getting(Jar::class) {
     attributes("Main-Class" to "app.cash.backfila.service.BackfilaDevelopmentServiceKt")
   }
   isZip64 = true
-  archiveClassifier.set("unshaded")
```
escardin, line 97: Right now this is technically breaking, as folks depending on the (unexpected) unshaded classifier will have broken builds. I'm working on keeping the unshaded jar, but I'm struggling somewhat with navigating gradle to that end.

adrw:
I'm still a bit lost why this is necessary or what it's fixing? Could you elaborate on what error you're hitting and how this fixes it? Are you having issues downloading and connecting sources in IntelliJ? Running the local development service?

escardin:
> I'm still a bit lost why this is necessary or what it's fixing? Could you elaborate on what error you're hitting and how this fixes it? Are you having issues downloading and connecting sources in IntelliJ? Running the local development service?

The issue is that it's ultimately not in line with how maven based dependencies are expected to work. By adding a classifier to the unshaded jar, tooling (including intellij) can't figure out how to associate the source and javadoc jars, because it sees them as fundamentally different.

I think this is a bit less of a problem with gradle, but with bazel it's just a huge pain to deal with.

The active, oh I really need to fix this was that because the source is kotlin, intellij wouldn't let me set breakpoints when trying to debug without associating the source code, which was deeply painful. This will fix that without breaking anyone.

adrw (approved):
LG

oldergod (approved):

PR #576, opened 2026-05-28
Depend on ReadyService and not SchemaMigrationService

escardin:
For self backfills, ReadyService is most appropriate, because backfila is actually registering itself, and therefore needs a working http api. Arguably it probably needs to wait for JettyService, but if SchemaMigrationService was sufficient, then ReadyService will also be fine.

RunnerSchedulerService is slightly more debatable. The lease hunter as designed expects a DB (which should be fully migrated), but in more general concept, one probably doesn't want to schedule runs or any of that until the service is 'ready'.

This is needed so I can use a different schema migrator than the built in one.


adrw (approved):
Looks good since SchemaMigratorService already is configured to be part of the Ready Service pre-reqs chain so this should be a no-op. https://github.com/cashapp/misk/blob/fe1a8b0b4039554e80fa9539de7bc889e4856797/misk-jdbc/src/main/kotlin/misk/jdbc/JdbcModule.kt#L143

PR #585, opened 2026-06-15
Add ABI compatibility checks to client libraries

keeferrourke:
## What

Adds Kotlin ABI/binary compatibility validation to the published backfila client libraries, mirroring the pattern used in [cashapp/misk](https://github.com/cashapp/misk).

## How

- Apply `org.jetbrains.kotlinx.binary-compatibility-validator` (0.18.1, the same version misk uses) at the root build via the version catalog, configuring every subproject automatically.
- Configure `apiValidation` with misk's defensive `ignoredProjects` pattern, excluding non-library modules: `bom`, `client-sqldelight-gradle-plugin`, `client-sqldelight-test`, `service`, and `service-self-backfill`.
- Commit each client library's public API baseline under `<module>/api/<module>.api` (generated via `gradle apiDump`). `build/api/` outputs remain gitignored.
- Add a dedicated `Binary Compatibility` CI job in a new `Linting.yml` workflow that runs `gradle apiCheck`, mirroring misk's `Linting.yml`. `apiCheck` is also wired into `gradle check`, so the existing Build workflow covers it too.
- Document the `gradle apiDump` workflow in `CONTRIBUTING.md`.

## Notes

- Three deprecated modules (`client-misk-dynamodb`, `client-misk-jooq`, `client-misk-static`) produce empty `.api` files — correct, since they contain only type aliases, which have no compiled ABI surface. They are still published, so they remain tracked.

## Verification

- `gradle apiDump` generated all 15 baselines successfully.
- `gradle apiCheck` passes.
- Confirmed the 5 ignored projects have no `apiCheck` task.

### DIFF
--- a/.github/workflows/Linting.yml
+++ b/.github/workflows/Linting.yml
@@ -1,0 +1,29 @@
+name: Linting
+
+on:
+  pull_request:
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+  push:
+    branches:
+      - master
+
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=3 -Xmx4096m -Dorg.gradle.configureondemand=true -Dkotlin.incremental=false -Dorg.gradle.jvmargs="-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8"
+  TERM: dumb
+
+jobs:
+  binaryCompatibility:
+    name: Binary Compatibility
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Hermit
+        run: ./bin/hermit env -r >> $GITHUB_ENV
+
+      - name: API Check
+        run: gradle apiCheck
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,9 +29,9 @@ buildscript {
 
 apply(plugin = "com.vanniktech.maven.publish.base")
 
-// ABI/binary compatibility validation for the published client libraries. The plugin is applied at
-// the root and automatically configures every subproject; we ignore the projects that aren't
-// consumable client libraries (the service, the BOM, the Gradle plugin, and test-only modules).
+// ABI/binary compatibility validation for the published libraries. The plugin is applied at the
+// root and automatically configures every subproject; we ignore the projects that don't publish a
+// consumable JVM API (the BOM, the Gradle plugin, and test-only modules).
 apply(plugin = "org.jetbrains.kotlinx.binary-compatibility-validator")
 
 configure<kotlinx.validation.ApiValidationExtension> {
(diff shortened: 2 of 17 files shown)


.github/workflows/Linting.yml
keeferrourke, line 1: Basically just copied from Misk

gradle/libs.versions.toml
```diff
@@ -3,6 +3,7 @@ jackson = "2.21.3"
 jetty = "10.0.26"
 jooq = "3.18.25"
 kotlin = "2.2.21"
+kotlinBinaryCompatibilityPlugin = "0.18.1"
```
keeferrourke, line 6: same as misk

build.gradle.kts
```diff
@@ -22,12 +22,31 @@ buildscript {
...
     classpath("app.cash.backfila:client-sqldelight-gradle-plugin")
   }
 }
 
 apply(plugin = "com.vanniktech.maven.publish.base")
 
+// ABI/binary compatibility validation for the published client libraries. The plugin is applied at
+// the root and automatically configures every subproject; we ignore the projects that aren't
+// consumable client libraries (the service, the BOM, the Gradle plugin, and test-only modules).
+apply(plugin = "org.jetbrains.kotlinx.binary-compatibility-validator")
```
keeferrourke, line 35: Backfila isn't using the `plugins` DSL yet

build.gradle.kts
```diff
@@ -22,12 +22,31 @@ buildscript {
...
+
+configure<kotlinx.validation.ApiValidationExtension> {
+  // Ignore subprojects only if present. This keeps the configuration robust if a project is
+  // removed or excluded via a settings overlay.
+  val ignorable = setOf(
+    "bom",
+    "client-sqldelight-gradle-plugin",
+    "client-sqldelight-test",
+    "service",
+    "service-self-backfill",
```
escardin, line 44:
```suggestion
    "client-sqldelight-test",
    "service-self-backfill",
```

Can you add service to that ABI? Backfila doesn't have the code required to run out of the box and the docker container doesn't work.

It's my hope to upstream something eventually, but in the meantime everyone depends on the service api staying stable.

keeferrourke: Can do

oldergod (approved):
Why is `service-self-backfill` excluded despite being a published JVM artifact?

Either we check it, either we stop publishing?

keeferrourke:
> Why is `service-self-backfill` excluded despite being a published JVM artifact?
> 
> Either we check it, either we stop publishing?

It seemed like a test lib so I instructed amp to omit it, but it actually is client-shaped despite not following the naming convention. I could add ABI checks for it, but it actually doesn't appear to be used by anything at Block (Square, Cash App) outside of our backfila wrapper service. Dunno about Nav and Faire.

If we want to stop publishing, maybe that should be a separate change, but it still doesn't seem like a good candidate to protect the ABI.

mpawliszyn:
service-self-backfill exists so backfila can backfill its own data

adrw (approved):
LG

keeferrourke:
> service-self-backfill exists so backfila can backfill its own data

Aha, then it should be published, but it seems less important to be ABI-protected.

PR #586, opened 2026-06-19
Report real DynamoDB scanned/matching record counts to metrics

sfavron:
## Summary

The DynamoDB operators (both v1 and v2) hardcode `matching_record_count(1L)` / `scanned_record_count(1L)` per segment-batch in `getNextBatchRange`. Because each batch maps to one segment, these are effectively segment-completion counters: the matching/scanned ratio is always ≈1, so it can't be used as a sparseness signal.

Computing real counts up front in `getNextBatchRange` would require an extra scan of each segment (~2x read capacity, which the client deliberately avoids). But the scan that already runs in `runBatch` returns the true counts (`ScanResponse.count()`/`scannedCount()` in v2; `ScanResultPage.count`/`scannedCount` in v1) **for free**. This change reports those actuals.

This is the "metrics-only" variant: the real counts feed the `run_batch_completed_records_{matching,scanned}_total` Prometheus metrics, giving a meaningful matching/scanned ratio. Persisted progress (`backfilled_*_record_count`), the rate counters, and the ETA are intentionally left on the segment-based `getNextBatchRange` estimate so they stay consistent with the segment-based `computed_*` precomputed totals — making the progress bar/ETA themselves record-accurate would additionally require count-ahead in `getNextBatchRange` (a separate, more expensive change).

### Changes
- **proto**: add optional `scanned_record_count` (4) / `matching_record_count` (5) to `RunBatchResponse`. Optional + new tag numbers = wire-compatible; clients that don't set them are unaffected.
- **client-dynamodb / client-dynamodb-v2**: accumulate the scan result's counts across the checkpoint loop and report them on the response.
- **service `BatchAwaiter`**: use the client's actuals for the completed-records metrics when present (per-RPC, including the partial `remaining_batch_range` path), falling back to the estimate otherwise.
- **backfila-embedded**: expose `runMatchingCount` / `runScannedCount` on `BackfillRun` so the real counts are observable in tests.

## Test plan
- New assertions in v1 + v2 `DynamoDbFilteringTest`:
  - No filter expression → 62 scanned / 62 matching (ratio 1.0, as expected when nothing is filtered).
  - DynamoDB filter expression → 62 scanned / 57 matching (the real sparseness signal).
- `BackfillRunnerTest` passes unchanged, exercising the estimate-fallback path for clients that don't report counts.
- `service` and `backfila-embedded` compile clean.

### DIFF
--- a/backfila-embedded/api/backfila-embedded.api
+++ b/backfila-embedded/api/backfila-embedded.api
@@ -29,6 +29,8 @@ public abstract interface class app/cash/backfila/embedded/BackfillRun {
 	public abstract fun getPrepareBackfillResponse ()Lapp/cash/backfila/protos/clientservice/PrepareBackfillResponse;
 	public abstract fun getRangeEnd ()Ljava/lang/String;
 	public abstract fun getRangeStart ()Ljava/lang/String;
+	public abstract fun getRunMatchingCount ()J
+	public abstract fun getRunScannedCount ()J
 	public abstract fun getScanSize ()J
 	public abstract fun partitionScan (Ljava/lang/String;)Lapp/cash/backfila/protos/clientservice/GetNextBatchRangeResponse;
 	public abstract fun precomputeRemaining ()V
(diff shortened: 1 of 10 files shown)


keeferrourke (approved):
Nice

PR #596, opened 2026-08-13
Expose backfill approval metadata

AaronVasquez:
## Why
Approval enforcement needs exact operator-visible evidence of the immutable registration policy and the persisted per-run approval. Without it, the dashboard APIs and audit event cannot distinguish pending from approved runs or identify which registration version governed a run.

## What
- Expose registration IDs and approval requirements in the registered-backfills response
- Expose the immutable registration snapshot and pending/approved fields in run status
- Preserve the pre-existing response constructors, data-class component/copy signatures, and JSON fields while adding the metadata getters
- Populate the existing audit approver LDAP field from persisted run approval on start
- Preserve the run creator as audit requestor for a newly approved start while retaining the actual operator in listener notifications and descriptions
- Prove re-registration does not change a run snapshot and both dry/wet approvals become visible

## Risk Assessment
Low — response fields are additive and audit output uses existing LDAP fields. The approval-aware path changes audit attribution only; enforcement, ordinary starts, and resumes retain their current behavior. The service API diff is additive-only, and focused serialization assertions cover the new fields.

## References
- Stacked on #595; review the 164-line child diff independently. This PR must merge after #595.
- Browser Approve-and-start is intentionally a separate follow-up slice.

## Update — 2026-08-12
- Added an action-level regression covering a run created by molly and approved by diana, plus the later resume-by-emma behavior.

Generated with Codex

PR #595, opened 2026-08-13
Enforce required backfill approvals

AaronVasquez:
## Why
`requires_approval` and per-run approval fields are persisted today, but transitions to `RUNNING` do not enforce them. Approval-required backfills can therefore start through the dashboard, direct state toggler calls, or create-and-start without a second human.

## What
- Add an explicit `approve` field to the existing authenticated start request
- Atomically persist a distinct human approver and start the run
- Gate every required transition to `RUNNING`, including dry runs, wet runs, resumes, and create-and-start
- Reject missing, partial, or same-creator approval state and keep clones unapproved

## Risk Assessment
Medium — this changes the central run-state transition, but only registrations with `requires_approval` enabled are newly blocked. Backfills that do not require approval retain their current behavior.

Generated with Codex

PR #597, opened 2026-08-13
Add browser approval flow for protected backfills

AaronVasquez:
## Summary

- replace browser mutation links with authenticated form-urlencoded POST actions guarded by exact same-origin Fetch Metadata
- show approve-and-start only to a distinct authenticated user and disable creator self-approval in the UI
- route start and approval through a dedicated endpoint, reject generic update bypasses, and preserve matching Turbo frames for success and error responses

## Stack

This is stacked on #596, which is stacked on #595. Merge in that order. The independent child diff is commit `4561ca91` (5 files, +260/-44).

## Risk

High. This adds the normal browser path for approving and starting protected Backfila runs. Backfila core remains the authoritative enforcement boundary; the UI is defense in depth and does not weaken direct-API enforcement.

## Verification

- `bin/gradle :service:test --tests app.cash.backfila.ui.actions.BackfillShowButtonHandlerActionTest :service:spotlessCheck :service:apiCheck :service:check`
- focused routed test: 5/5 passed
- GET start is not routed; cross-site, same-site, `none`, and missing Fetch Metadata are rejected with 403
- same-origin backend failure returns 422 with the matching Turbo frame
- generated `service.api` is byte-identical
- `git diff --check`

## Follow-up gate

Cash-server must add this dedicated `POST /api/backfill/{id}/start` route to its authorization contract before the Backfila artifact is deployed. No Backfila run or production mutation is part of this PR.


AaronVasquez:
🤖 @abchouti @clararende @kazemek21 — the fork PR cannot create formal upstream review requests, so tagging you here for review. This is the browser approval/CSRF follow-up stacked on #596; the independent child diff is commit `4561ca91`.

PR #599, opened 2026-08-31
Fix sparse-predicate batching in Jooq

hugomd:
## Summary

Fixes sparse-predicate batching in the jOOQ client by bounding work with an unfiltered raw-key window of at most `scan_size` rows before applying the backfill predicate.

## Problem

Prepared jOOQ backfills normally provide an overall range end. `OpenKeyRange` previously used that value directly, bypassing its `LIMIT scan_size` path.

`BatchRangeIterator` would then:

1. Find up to `batch_size` matching keys.
2. Use the first and last matches as the batch range.
3. **Run an unfiltered `COUNT(*)` across that raw-key interval.** This times out for large tables.

For sparse predicates, a few matching rows can span millions of raw IDs. Although `LIMIT batch_size` bounds matching rows, it does not bound the subsequent count.

## Approach

The new batching flow mirrors Hibernate Backfila’s two-level model:

1. Select the next `scan_size` raw keys without applying the backfill predicate, capped by the overall range end.
2. Apply the predicate only within that raw-key window.
3. End a full batch at its `batch_size`th match.
4. Extend partial or empty batches to the raw window’s upper bound.
5. Continue within the same window when a full batch ends early; otherwise advance to a new raw window.

This guarantees every emitted raw range contains at most `scan_size` rows, including windows where no rows match.

For example, with `scan_size = 4`, `batch_size = 2`, and matches at positions `0, 1, 8, 11`:

| Batch | Raw range | Scanned | Matching |
|---|---:|---:|---:|
| Full batch | `0..1` | 2 | 2 |
| Empty window remainder | `2..3` | 2 | 0 |
| Entirely empty window | `4..7` | 4 | 0 |
| Sparse full batch | `8..11` | 4 | 2 |

`scan_size` is a ceiling rather than a target: a batch can end earlier when it reaches `batch_size` matches.

## Implementation details

- Builds each raw window using an ordered, unfiltered `LIMIT scan_size` query.
- Remaps compound-key fields to the derived scan-window table before selecting its maximum key.
- Determines the first real raw key separately so inclusive batch ranges do not replay the preceding match.
- Defers computing the next raw window until another batch is requested, avoiding unnecessary queries when the compute limit is reached.
- Preserves the existing public `OpenKeyRange` API and precomputation behavior.

## Regression coverage

Adds scalar- and compound-key sparse-predicate tests that verify:

- No emitted raw range exceeds `scan_size`.
- Empty raw windows and empty window tails advance correctly.
- Matching rows are processed exactly once without skips or duplicates.
- Compound-key ordering works when the first two key fields tie and the binary final field determines ordering.

### DIFF
--- a/client-jooq/src/main/kotlin/app/cash/backfila/client/jooq/internal/BatchRangeIterator.kt
+++ b/client-jooq/src/main/kotlin/app/cash/backfila/client/jooq/internal/BatchRangeIterator.kt
@@ -21,11 +21,12 @@ class BatchRangeIterator<K, Param : Any>(
 ) : AbstractIterator<GetNextBatchRangeResponse.Batch>() {
 
   private val timeElapsed: () -> Boolean
-  private var nextKeyRange: Lazy<OpenKeyRange<K>>
+  private var nextKeyRange: OpenKeyRange<K>
+  private var pendingWindow: PendingWindow<K>? = null
   init {
     timeElapsed =
       if (request.compute_time_limit_ms == null) { { false } } else timer(request.compute_time_limit_ms)
-    nextKeyRange = lazy { OpenKeyRange.initialRangeFor(jooqBackfill, request, session) }
+    nextKeyRange = OpenKeyRange.initialRangeFor(jooqBackfill, request, session)
   }
 
   private fun timer(timeLimitMs: Long): () -> Boolean {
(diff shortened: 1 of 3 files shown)


rgrnwd (approved):
Maybe add a feature flag to roll this out safely?

oldergod (approved):

adrw (approved):

frojasg (approved):

PR #600, opened 2026-08-31
fix(service): key Edit Cursor on partition id so it works for any partition name

guodong-sq:
### Background

The "Edit Cursor" link on the backfill page has never worked for a partition whose name needs
percent-encoding in a URL path. It has been broken this way since #448 introduced it.

`EditPartitionCursorAction` builds its own URL by substituting into a path template:

```kotlin
private const val PATH = "/backfills/{id}/{partitions}/{partitionName}/edit-cursor"
fun path(id: Long, partitionName: String) = PATH
  .replace("{id}", id.toString())
  .replace("{partitionName}", partitionName)
```

Two defects on those four lines, plus one in misk's binding, stack up:

1. `{partitions}` is never substituted, so every rendered href carries that literal text. misk treats
   the unbound template variable as a wildcard segment, so the route still resolves and the typo is
   invisible.
2. `partitionName` is interpolated raw. A name containing `#` truncates the URL at the fragment
   boundary, so the browser requests a shorter path that matches no route.
3. misk matches routes against `HttpUrl.encodedPath` (`PathPattern.matcher`) and binds `@PathParam`
   through an identity `StringConverter`, so a correctly percent-encoded name arrives at the action
   still encoded and `partitions.find { it.name == partitionName }` cannot match it. Encoding the
   link alone does not fix this; the action would also have to decode.

Reproduced on https://backfila.stage.sqprod.co/backfills/24852 (Block-internal staging), an `orders`
backfill whose partitions are named `region-shard-us-east#orders_local_ro_0`:

```
rendered href  /backfills/1/{partitions}/region-shard-us-east#orders_local_ro_0/edit-cursor
browser GET    /backfills/1/%7Bpartitions%7D/region-shard-us-east                      404
encoded GET    /backfills/1/%7Bpartitions%7D/region-shard-us-east%23orders_...          500
handler GET    /backfills/1/region-shard-us-east%23orders_.../edit-cursor               200
               "Edit partition failed. Partition not found: region-shard-us-east%23orders_local_ro_0"
```

Both failures are invisible to the operator. The link sits inside the partitions `<turbo-frame>`, and
misk returns `text/plain` for 404 and for "internal server error", which Turbo's `loadResponse` drops
without rendering. The button appears to do nothing at all.

Partition names come from client services verbatim and are never sanitized, and two bundled clients
already emit names that break the same link: the DynamoDB operators use `"$i of $partitionCount"`
(spaces) and the S3 operator uses object key suffixes (slashes).

### Change(s) Made

1. Address the partition by its numeric id instead of its name. `UiPartition.id` was already exposed
   and the write already keyed on it, so this removes the string from the URL entirely. No character
   class can break the link, and there is no decode for a future change to forget.
2. Give the two actions distinct, fully substituted paths: `/backfills/{id}/partitions/{partitionId}/edit-cursor`
   for the form, and `/api/backfill/{id}/partitions/{partitionId}/cursor` for the handler, matching
   where `BackfillShowButtonHandlerAction` puts its mutating endpoint. Simply deleting the stray
   `{partitions}` segment would have made the two `@Get` templates identical.
3. Retype `RunPartitionQuery.partitionId` from `Long` to `Id<DbRunPartition>`, matching every other
   constraint in that interface and preventing a backfill run id from being passed where a partition
   id is expected. This is a type-consistency change and not a correctness one: I checked, and the
   raw `Long` form binds and executes correctly. Happy to drop it if you would rather keep the diff
   and the ABI baseline narrower.
4. Read the form values with `@QueryParam` instead of unwrapping `ActionScoped<HttpCall>` by hand,
   and require a non-blank `new_cursor`. Previously a request without that parameter set the cursor
   to `NULL`, restarting the partition from the beginning of its range; `required = true` on the
   input was the only guard and it is client side only. This also replaces `isValidUtf8`, which
   compared `input.toByteArray(UTF_8)` to itself and could never return false.
5. Move the snapshot comparison into the write transaction and compare `ByteString` to `ByteString`.
   The snapshot was previously read in one transaction and applied in another, and it compared the
   lossy `pkey_cursor?.utf8()` rendering, so a cursor holding non UTF-8 bytes would pass validation
   and then be overwritten with the UTF-8 encoding of its own mangled rendering. That mattered more
   than it looks, because the form pre-fills `new_cursor` with the same lossy rendering, so
   submitting the form unedited silently rewrote the bytes.
6. Because of (5), a cursor whose bytes are not UTF-8 can no longer round-trip through the form at
   all. Rather than report that as a phantom concurrent modification, the handler now distinguishes
   the two refusals and says the cursor is not valid UTF-8, so nobody chases a writer that does not
   exist. `client-jooq` publishes a `forByteArray` key serializer, so this is a reachable state
   rather than a hypothetical one.

`Response` for a failed edit still carries status 200, and the mutation is still an `@Get`. Both are
pre-existing and shared with the sibling handler action, so they are left alone here.

### Scope & Impact

Backfila dashboard only. Two route paths change; neither previously resolved to a working page for
any name requiring encoding, and the form path never resolved at all because `path()` emitted a
literal `{partitions}`, so there are no working URLs to keep compatible.

`service/api/service.api` is regenerated per CONTRIBUTING.md. The changes are confined to the
`service` module, which is not a published client library, so no client ABI moves.

Behaviour change worth calling out: clearing a cursor back to `NULL` through this endpoint is no
longer possible. It was only reachable by hand editing the URL, and it deserves its own explicit
affordance if anyone wants it.

### Related Documents/PRs

1. #448, which introduced the feature. Its only test was removed during review before merge, so
   there has been no coverage of either action.

### Testing Performed

New `service/src/test/kotlin/app/cash/backfila/actions/EditPartitionCursorActionTest.kt`, eight tests:

1. `paths never interpolate the partition name` pins both generated paths against literals. Against
   the old code it neither compiles (the old `path()` took the name as a `String`) nor passes (it
   returned a literal `{partitions}` segment). This is the assertion that pins the bug class.
2. `form page renders for a partition whose name is not url safe` covers the form action, including
   its `BadRequestException` on an unknown partition id. Note `buildHtmlResponseBody` defers to
   `writeTo`, so this asserts the action completes rather than asserting on rendered markup.
3. `edits a partition whose name is not url safe` and
   `edits a partition using the matching non null snapshot the form submits` cover both branches of
   the rewritten comparison. Both pass `cursor_snapshot` as the form actually sends it, empty for a
   never-started partition rather than absent.
4. `rejects a stale snapshot`, `a missing new cursor does not clear the cursor`, and
   `refuses a partition belonging to a different backfill` pin the guards, the last one covering the
   new `backfillRunId` scoping on the write query.
5. `leaves a cursor that is not utf8 untouched` writes raw `FF FE` to `pkey_cursor` and asserts the
   bytes survive the attempted edit. This fails on master, which overwrote them with `EF BF BD`.

These call the actions directly rather than over HTTP, so they cover the write path rather than URL parsing;

### DIFF
--- a/service/api/service.api
+++ b/service/api/service.api
@@ -201,21 +201,22 @@ public final class app/cash/backfila/dashboard/CreateBackfillAction : misk/web/a
 public final class app/cash/backfila/dashboard/EditPartitionCursorAction : misk/web/actions/WebAction {
 	public static final field Companion Lapp/cash/backfila/dashboard/EditPartitionCursorAction$Companion;
 	public fun <init> (Lapp/cash/backfila/dashboard/GetBackfillStatusAction;Lapp/cash/backfila/ui/components/DashboardPageLayout;)V
-	public final fun get (JLjava/lang/String;)Lmisk/web/Response;
+	public final fun get (JJ)Lmisk/web/Response;
 }
 
 public final class app/cash/backfila/dashboard/EditPartitionCursorAction$Companion {
-	public final fun path (JLjava/lang/String;)Ljava/lang/String;
+	public final fun path (JJ)Ljava/lang/String;
 }
 
 public final class app/cash/backfila/dashboard/EditPartitionCursorHandlerAction : misk/web/actions/WebAction {
 	public static final field Companion Lapp/cash/backfila/dashboard/EditPartitionCursorHandlerAction$Companion;
-	public fun <init> (Lapp/cash/backfila/dashboard/GetBackfillStatusAction;Lmisk/hibernate/Transacter;Lmisk/hibernate/Query$Factory;Lmisk/scope/ActionScoped;Lapp/cash/backfila/ui/components/DashboardPageLayout;)V
-	public final fun get (JLjava/lang/String;)Lmisk/web/Response;
+	public fun <init> (Lapp/cash/backfila/dashboard/GetBackfillStatusAction;Lmisk/hibernate/Transacter;Lmisk/hibernate/Query$Factory;Lapp/cash/backfila/ui/components/DashboardPageLayout;)V
+	public final fun get (JJLjava/lang/String;Ljava/lang/String;)Lmisk/web/Response;
+	public static synthetic fun get$default (Lapp/cash/backfila/dashboard/EditPartitionCursorHandlerAction;JJLjava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Lmisk/web/Response;
 }
 
 public final class app/cash/backfila/dashboard/EditPartitionCursorHandlerAction$Companion {
-	public final fun path (JLjava/lang/String;)Ljava/lang/String;
+	public final fun path (JJ)Ljava/lang/String;
 }
 
 public final class app/cash/backfila/dashboard/GetBackfillRunsAction : misk/web/actions/WebAction {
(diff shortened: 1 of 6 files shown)


guodong-sq:
@buiderbot review

cemcetin (approved):

frojasg (approved):

PR #604, opened 2026-09-11
Add generic SQLDelight query interception

eeSeeGee:
## Why

Some SQLDelight Backfila record sources need scoped changes to query execution. SQLDelight does not expose generated SQL for callers to decorate, and putting database-specific behavior in the Gradle plugin couples general code to one backend.

## What

This adds a generic query interceptor to SQLDelight Backfila and a scoped `SqlDriver` decorator that transforms SQL only while one Backfila query executes.

Existing backfills remain unchanged by default.

## How

`SqlDelightDatasourceBackfill` accepts an optional `SqlDelightQueryInterceptor`. The record source routes each generated query execution through it.

`ScopedSqlDriverInterceptor` implements both the interceptor and `SqlDriver`. A caller builds its SQLDelight database with this driver, then passes the same instance only to backfills that need the transform. The decorator preserves generated binders and row mappers.

The scope uses thread-local state and supports nested calls. It is documented for synchronous drivers whose execution stays on the calling thread.

The Gradle plugin and generated record-source configs contain no Vitess-specific behavior.

## Risk

Low for existing users. The original one-argument constructor remains available, and the default interceptor is a no-op. New callers control the SQL transform and must ensure it is valid for their database.

## Testing

No manual testing. CI covers the scoped driver behavior and binary compatibility.

## Bigger picture

Contacts can use this hook to attach `ALLOW_SCATTER` only to intentional SQLDelight Backfila scans while keeping scatter enforcement enabled elsewhere.

Generated with Goose


client-sqldelight-gradle-plugin/src/main/kotlin/app/cash/backfila/client/sqldelight/plugin/BackfilaSqlDelightGradlePlugin.kt
```diff
@@ -100,6 +100,7 @@ abstract class BackfilaSqlDelightExtension {
     recordColumns: String,
     recordType: String,
     whereClause: String = "1 = 1",
+    allowScatter: Boolean = false,
```
mpawliszyn, line 103: I'd rather not let vitess internals leak into the general SqlDelight gradle plugin

client-sqldelight/src/main/kotlin/app/cash/backfila/client/sqldelight/internal/SqlDelightRecordSource.kt
```diff
@@ -13,6 +13,9 @@ class SqlDelightRecordSource<K : Any, R : Any>(
 ) {
   val keyEncoder = recordSourceQueries.keyEncoder
 
+  private fun <T> execute(block: () -> T): T =
+    recordSourceQueries.allowScatterDriver?.allowingScatter(block) ?: block()
```
mpawliszyn, line 17: Maybe we can make this interceptable and we get this by intercepting with the driver or some other component instead?

mpawliszyn (approved):
Much better!

